PR #15, opened 2015-03-30
Fix cofor so the samples run again

tommyettinger:
The implementation of cofor has an i (unbound variable name) where it should probably have a _. That should be the only change that I put in this PR.


aardappel:
Whoops, thanks!

PR #18, opened 2016-04-10
Some fixes for the gcc build

binji:
* INT_{MIN,MAX} are defined in limits.h
* workaround a nasty template parsing bug in g++:
  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=10200


aardappel:
Wow, that's a crazy bug, how did you find it?

binji:
Well, it wouldn't compile, but the code looked good. So I tried building with clang and it worked... so then I searched around. :)

aardappel:
Reported in 2003, fix will come for GCC 6. Wow.

PR #29, opened 2019-04-04
Update shooter tutorial to use let/var/class syntax

joshpurvis:
I was working through the shooter tutorial as a learning exercise, and realized it was using `:=`, `:==`, and `struct` which seem to no longer be part of the language. I've changed all these instances to `let`, `var`, and `class` respectively to get it running again.

Also there was another minor mistake in the tutorial code -- the player/enemy collision detection line was referencing a variable that wasn't previously declared. I changed it to:
```
if magnitude(playerpos - e.pos) < 1:
    # ...
```

There are some other files in the docs directory which will need to be changed as well.

### DIFF
--- a/lobster/docs/shooter_tutorial.html
+++ b/lobster/docs/shooter_tutorial.html
@@ -49,7 +49,7 @@ include color
 
 fatal(gl_window(&quot;Shooter Tutorial&quot;, 640, 480))
 
-worldsize := 20.0
+let worldsize = 20.0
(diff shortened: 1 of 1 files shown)


aardappel:
Thanks! I think I updated it in https://github.com/aardappel/lobster/blob/master/lobster/docs/source/shooter_tutorial.md but failed to also update the html files (there's a script for that)

PR #32, opened 2019-04-06
Updated c-style cheat sheet

kyoto-shift:
I think Lobster is a really great language/idea, so I wanted to contribute any way I could. To also learn how Lobster works, I wrote a basic doc generator (docs/source/generate_docs.lobster) so I could easily update the c-style cheat sheet. I tested each example in the cheat sheet (tests/cstylecheatsheettest.lobster), updated any that needed to be (e.g. usage of ":="), and removed any that no longer worked (dynamic scoping, etc.).


aardappel:
Hah, that's a really crazy project! It even generates almost the same html as before, impressive!

That said, from a practical POV, I am not sure we want to introduce an entirely new document format and processor for the sake of one `.html` file. If anything, the html should be translated to `.md`, I convert them with Pandoc which should be able to do tables. Not sure why I made this document in html, I guess I couldn't get the layout quite right in `.md` at the time.

kyoto-shift:
> That said, from a practical POV, I am not sure we want to introduce an entirely new document format and processor for the sake of one `.html` file. If anything, the html should be translated to `.md`, I convert them with Pandoc which should be able to do tables. Not sure why I made this document in html, I guess I couldn't get the layout quite right in `.md` at the time.

Oh right, I forgot to mention! My original idea was to convert the cheat sheet to .md, but while I was rewriting it, I was having to rely on some odd "hack-y" solutions (manually fixing column width with `&nbsp;` characters) to get tables to look like they did in the original. The styling definitely looked better, but I didn't feel comfortable contributing it if I was using "hack-y" methods. I only really intended the doc tool to be used for the cheat sheet or any other comparison-like documents. I definitely agree that it's not very practical, though.

Should I rewrite the cheat sheet in markdown using the solutions I found instead?

aardappel:
Yes, sorry for the effort you put in the converter, but I think that would be better in the long run. I think you had fun though? :)

Maybe you can first give an example of what the markdown would look like?

kyoto-shift:
No worries! It was actually a ton of fun.


So here's a small example of what the tables would look like:

| Lobster Version                                                                          | C-Style Version                                                                                                                                             | Notes |
| --------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- | ----- |
| if a: b else: c                                                                         | if (a) b; else c                                                                                                                                            |       |
| if a:<br/>&nbsp;&nbsp;b<br/>&nbsp;&nbsp;c<br/>else:<br/>&nbsp;&nbsp;d<br/>&nbsp;&nbsp;e | if&nbsp;(a)&nbsp;{<br />&nbsp;&nbsp;b&semi;<br />&nbsp;&nbsp;c&semi;<br />}&nbsp;else&nbsp;{<br />&nbsp;&nbsp;d&semi;<br />&nbsp;&nbsp;e&semi;<br />}<br /> |       |
| [...]                                                                                   | [...]                                                                                                                                                       | [...] |
| let&nbsp;r&nbsp;=&nbsp;map(list)&nbsp;x:&nbsp;x&nbsp;&ast;&nbsp;x<br />                 | var&nbsp;r&nbsp;=&nbsp;list.ConvertAll(x&nbsp;=&gt;&nbsp;x&nbsp;&ast;&nbsp;x)&semi;<br />                                                                   | Most modern programming languages are able to use some form of lambdas that are similar, but not quite as powerful as the ones in Lobster (e.g., not being able to use `return` to break out of a loop).      |

*(the font of the code would most likely be monospace in the actual file)*

To make the code examples properly fit into a table row, I had to write a small tool for character conversion since you can't (easily) have line breaks or certain characters in a table row without breaking the structure and relying on "hack-y" solutions.

The tool just takes something like: ` let r = map(list) x: x * x `  and converts it into: `let&nbsp;r&nbsp;=&nbsp;map(list)&nbsp;x:&nbsp;x&nbsp;&ast;&nbsp;x<br />` so it can be put directly into a table row and not have to abuse any weird quirks of markdown. While this works, it's also not super practical, but is better than manually trying to write code to fit in a table row. [Here's the source for that tool](https://gist.github.com/kyoto-shift/726b366a05c4aa1537af7b911db6601b). 

What do you think?

aardappel:
Ok, that's a small little converter we can deal with :) Does that run on the md or the html? I presume the md?

kyoto-shift:
You're correct! And awesome, I'll finish updating it then.

kyoto-shift:
All done! I slightly rewrote the converter (`docs/source/lobster_to_md.lobster`) so it's not limited to just one example per run and can also directly output markdown table rows now.

lobster/docs/source/lobster_to_md.lobster
```diff
@@ -0,0 +1,75 @@
+/*
```
aardappel: I still don't understand how this works. This code appears to work on an internal example, but where is the code from which `c_style_cheat_sheet.md` was generated?
Shouldn't this code be set up such that it generates `c_style_cheat_sheet.md`, and maybe call that from the script that generates all documents?

lobster/tests/cstylecheatsheettest.lobster
```diff
@@ -0,0 +1,81 @@
+import std
```
aardappel: I don't think we need this file in here, generating `c_style_cheat_sheet.md` would be enough of a test already?

lobster/docs/source/lobster_to_md.lobster
kyoto-shift: Ah, I think there might've been some misunderstanding on my part. I wrote `c_style_cheat_sheet.md` manually, just using the converter on code examples, then copying and pasting them into my editor which made sure they were properly formatted into a table. If you turn word wrapping off on that file, you'll see it's just a formatted table, not the mess word wrapping turns it into haha.

What would be the best way to generate such a file like `c_style_cheat_sheet.md`? Would it be something like my `.doc` syntax that&mdash;instead of HTML&mdash;outputs a markdown file, or something entirely different?

lobster/tests/cstylecheatsheettest.lobster
kyoto-shift: Since `c_style_cheat_sheet.md` isn't actually generated (yet?), I just included it so you or anyone else could see that the examples in the cheat sheet actually do work and were tested.

lobster/docs/source/lobster_to_md.lobster
aardappel: Ok, I am just thinking in terms of how this file should be edited in the future. All the `&nbsp;` in particular makes the code unreadable in the `.md`. And there is no easy way to edit this code in this form, or even to paste it into your tool.

Does the escaping even need to be this heavy handed, other than lt/gt?

I would love to have everything in `.md`, but if it becomes harder to edit than `.html`, that's not good. Also, needing a relatively complicated tool that understands most of `.md` doesn't sound great either.

kyoto-shift: Yeah, that's what I've been thinking about too. It's a weirdly specific problem to fix.

As for the characters that **need** to be converted. I'd say the most minimal set would be:
1. `<` and `>` -> `&lt;`, `&gt;`
2. `\n` -> `<br/>` &mdash; Since the code needs to fit in a markdown table row, we need some way to keep the original line breaks intact.
3. `\s` -> `&nbsp;` &mdash; When `<br/>` is used, there needs to be a way to keep the original indentation intact.

As long as the examples are wrapped in `<code>` tags, those should be the only characters that need to be converted (AFAIK).

aardappel: I was looking here: https://pandoc.org/MANUAL.html
Search for `grid_tables`.
Says: "The cells of grid tables may contain arbitrary block elements (multiple paragraphs, code blocks, lists, etc.). " which sounds promising
So in theory multiple lines of a cell could all preceded by 4 spaces to indicate a code block and be formatted correctly?

aardappel: Do you think using these tables is feasible?

aardappel:
Hey, shame this never went in. If can make this into readable markdown, we can re-open. Note that original files have changed by now.

PR #33, opened 2019-04-07
Add glTF functionality

fluffrabbit:
I guess the party starts here. In its current state, this will compile with cmake. As for how well it handles glTF, dunno.

There are likely shader pipeline mismatches. It also currently lacks animation support. There is an example in samples/gltftest.


dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+// OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+// ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+// OTHER DEALINGS IN THE SOFTWARE.
+
+#pragma once
+#ifndef FLAT_GLTF_H
+#define FLAT_GLTF_H
+
+#ifndef FLATBUFFERS_IDL_H_
+#error Include flatbuffers/idl.h before flat_gltf.h so glTF models can be loaded.
```
aardappel: Just include it here? If you're going with the style of including things in includes, might as well go all the way. Or alternatively, also remove the other includes below, and rely on the .cpp including stdafx.h before your include, like all other code does.

dev/src/gltfwrapper.cpp
```diff
@@ -0,0 +1,81 @@
...
+  nfr("gl_new_scene_gltf", "filename", "S", "R?",
+      "load a .gltf file into a scene, returns scene or nil on failure to "
+      "load.",
+      [](VM &vm, Value &fn) {
+        TestGL(vm);
+        auto s = LoadGLTF(fn.sval()->strv());
+        return s ? Value(vm.NewResource(s, &scene_type)) : Value();
+      });
+
+  nfr("gl_render_scene", "s", "R", "", "renders the specified scene",
```
aardappel: lets have both functions have the world `gltf` in them somewhere for clarity.

lobster/schemas/gltf.fbs
```diff
@@ -0,0 +1,130 @@
...
+  materials:[Material];
+  meshes:[Mesh];
+  textures:[Texture];
+  images:[Image];
+  accessors:[Accessor];
+  bufferViews:[BufferView];
+  buffers:[Buffer];
+}
+
+root_type GLTF;
```
aardappel: add a `file_extension gltfb` or whatever (glft binary?) here, that way `flatc` will save things with that extension rather than `bin`.. we are inventing a new file format here afterall.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+#endif
+
+#include <functional>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "glincludes.h"
```
aardappel: Any reason this is header only? I'd prefer this as `glloadgltf.cpp` so it fits in with `glloadiqm.cpp` and the rest of the gl*.cpp` files. That would also simplify the question of where headers go :)

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+extern Shader *currentshader;
+
+namespace fgl {
+
+bool verbose = true;
+
+#ifndef _VERBOUT_
+#define _VERBOUT_                                                              \
+  if (verbose)                                                                 \
+  std::cout
```
aardappel: We're not using `std::cout` anywhere in the lobster codebase. Please use LOG_ERROR, LOG_INFO and friends.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+#endif
+
+typedef struct {
+  bool success;
+  GLuint texture;
+  GLsizei width;
+  GLsizei height;
+  unsigned int channels;
+  bool mipmap;
+} Texture;
```
aardappel: This is C++, we can use `struct Texture {..};` instead.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+
+Texture newTexture = {false, 0, 0, 0, 0, false};
+
+Texture drawTexture0 = {false, 0, 0, 0, 0, false};
+
+float4x4 texMatrix = float4x4_1;
+
+// TODO: pbr_anim shader
+typedef struct {
+  GLfloat Position[4]; // vec4 position is not necessary; GLSL auto-converts vec3 to vec4
```
aardappel: can you use types like `float4` where possible.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  {1.0f, 1.0f, 1.0f, 1.0f},
+  {0.0f, 0.0f, 0.0f, 0.0f},
+  {0.0f, 0.0f, 0.0f, 0.0f}
+};
+
+typedef GLuint Index;
+
+typedef struct {
+  bool success;
+  GLuint vao[3]; // VAO state, vertex buffer, index buffer
```
aardappel: use 3 actual variables?

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  // upload the vertices, passing the size in bytes
+  glBufferData(GL_ARRAY_BUFFER, vertices.size() * sizeof(Vertex), &vertices[0],
+               streaming ? GL_STREAM_DRAW : GL_STATIC_DRAW);
+  _VERBOUT_ << "Number of vertices: " << vertices.size() << std::endl;
+  _VERBOUT_ << "Size of Vertex:     " << sizeof(Vertex) << std::endl;
+
+  // perhaps an inconsistent use of data here but I'm not sure what to do
+  // sizeof returns a size_t which can be cast to anything since it's just a
+  // number
+  std::size_t pointer = 0;
```
aardappel: Ah apologies, I never added a clang format file to this repo, and you formatted with the default style. I just added such a file. Can you run it again after you rebased against master?

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  // upload the indices, passing the size in bytes
+  glBufferData(GL_ELEMENT_ARRAY_BUFFER, indices.size() * sizeof(Index),
+               &indices[0], streaming ? GL_STREAM_DRAW : GL_STATIC_DRAW);
+  _VERBOUT_ << "Number of indices: " << indices.size() << std::endl;
+  _VERBOUT_ << "Size of Index:     " << sizeof(Index) << std::endl << std::endl;
+
+  // switch out of the VAO
+  glBindVertexArray(0);
+
+  // failure conditions are currently unimplemented
```
aardappel: The style of the rest of the codebase is not to use quite as many blank lines as you do. Also start comments with a captital and end with a `.`, i.e. a sentence.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  mesh.indexCount = indices.size();
+
+  return mesh;
+}
+
+void DrawMesh(Mesh &mesh, float4x4 modelViewMat, float4x4 projMat) {
+  if (!mesh.success)
+    return;
+
+  float4x4 mvp = modelViewMat * projMat;
```
aardappel: snake_case please

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+
+  // the casting of each field means double4x4 could potentially be used instead
+  // TODO FEATURE: no model matrix means no normal matrix, but sacrifices must
+  // be made :/
+  const GLfloat glmats[16] = {
```
aardappel: This is not needed, just get `mvp.data()`

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  glUniformMatrix4fv(currentshader->mvp_i, 1, GL_FALSE, glmats);
+
+  // switch graphics memory to the VAO
+  glBindVertexArray(mesh.vao[0]);
+
+  glDrawElements(GL_TRIANGLES, mesh.indexCount, GL_UNSIGNED_INT,
+                 (const GLvoid *)0);
+}
+
+Texture LoadTexture(const GLvoid *data, GLsizei width, GLsizei height,
```
aardappel: If whoever is calling this could just call `CreateTexture` from `glinterface.h`, that would be better. Generally, the more types and functions you can use from there, the better. If those don't implement what you need, feel free to modify them.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+
+void SetSceneAnimation(Scene &scene, int anim, double animTime) {
+  SceneAnimation &animation = scene.animations[anim];
+}
+
+void DrawScene(Scene scene, float4x4 modelViewMat, float4x4 projMat) {
+  size_t ct = 0;
+  int currentMaterial = -1;
+  // https://shaharmike.com/cpp/lambdas-and-functions/
+  std::function<void(int, float4x4)> DrawTree = [&](int n, float4x4 m) {
```
aardappel: use `auto`

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+    ct++;
+    if (ct > MAX_SCENE_NODES)
```
aardappel: How can this happen? Wouldn't we run out of memory for the file we're trying to load before hitting this node count? Maybe remove it or check this in another way?

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+      throw std::out_of_range(
```
aardappel: If at all possible, have functions return a bool, error code or error string to indicate a problem, so the caller can decide if a bad file should mean the program terminates. Prefer to not use exception handling.. I know the rest of the codebase does for fatal runtime errors, but I am actually trying to use it as little as possible.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+#define _ERROUT_ std::cerr
+#endif
+
+#ifndef MAX_SCENE_NODES
+#define MAX_SCENE_NODES 1000000
+#endif
+
+typedef struct {
+  int baseColorImage;
+  GLfloat metallicFactor;
```
aardappel: also snake_case for instance variables.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+Scene newScene = {{}, {}, {}, {}, {}};
+
+typedef struct {
+  std::string path;
+  fgl::Texture texture;
+} Image;
+
+std::vector<Image> images;
+
+int LoadImage(std::string gltfPath, std::string imgName, bool filter = true) {
```
aardappel: Can this be replaced with `CreateTextureFromFile` ?

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  unsigned char *imgData =
+      stbi_load_from_memory((stbi_uc*)fileData.data(), fileData.size(), &imgWidth, &imgHeight, &imgChannels, 0);
+  fgl::Texture imgTexture =
+      fgl::LoadTexture(imgData, imgWidth, imgHeight, imgChannels, filter, filter);
+  stbi_image_free(imgData);
+  images.push_back({imgFilePath, imgTexture});
+  return images.size() - 1;
+}
+
+Scene LoadScene(std::string gltfPath, std::string schemaPath = "gltf.fbs") {
```
aardappel: `string_view`, baby!

aardappel: Make the user specify where the schema is, i.e. all the way down to `gl_new_scene_gltf`

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+  _VERBOUT_ << "GLTF SCENE" << std::endl;
+  try {
+    std::string gltfText;
+    if (LoadFile(gltfPath.c_str(), &gltfText) < 0) {
+      _ERROUT_ << "Failed to open " << gltfPath << std::endl << std::endl;
+      return newScene;
+    }
+
+    std::string schemaText;
+    if (LoadFile(schemaPath.c_str(), &schemaText) < 0) {
```
aardappel: by default, code should load the binary version, that does not require the schema or json parsing. You can then have a utility function that loads it via json + schema. In fact, this utility already exists, and is `flatbuffers_json_to_binary` in file.cpp. You can also leave the file loading to the user (see `read_file` in file.cpp), that way this function that just receive the FlatBuffer binary as a `string_view` (taken from the input LString::strv()`). That way they can choose if they want to load binary directly or go via json.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+                break;
+              case 2: // tangents are not supported by the phong_anim shader
+                //verts[i].Tangent[0] = vDat[0];
+                //verts[i].Tangent[1] = vDat[1];
+                //verts[i].Tangent[2] = vDat[2];
+                break;
+              case 3:
+                verts[i].UV[0] = vDat[0];
+                // OpenGL UVs are vertically flipped, glTF UVs work like Vulkan UVs
+                verts[i].UV[1] = vDat[1] * -1.0f + 1.0f;
```
aardappel: 1 - vDat[1] ? :)

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+                &buffers[bv.buffer][bv.offset + a.offset]);
+            _VERBOUT_ << "Buffer: " << bv.buffer << std::endl;
+            _VERBOUT_ << "Byte offset: " << (bv.offset + a.offset) << std::endl;
+            _VERBOUT_ << "Byte stride: " << bStride << std::endl;
+            _VERBOUT_ << count << " vertex attribute vectors" << std::endl;
+            for (int i = 0; i < count; i++) {
+              GLfloat *vDat = &vBuf[bStride / 4 * i];
+              switch (atType) {
+              case 0:
+                verts[i].Position[0] = vDat[0];
```
aardappel: Just cast directly to `float3 *` or whatever.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+              // 3 components * 4 bytes
+              bStride = 12;
+            } else if (bStride == 0 && a.type == "VEC2") {
+              // 2 components * 4 bytes
+              bStride = 8;
+            }
+            // we only use floats for vertex attributes
+            if (a.componentType != 5126)
+              count = 0;
+            if (count > (int)verts.size())
```
aardappel: make `count` a `size_t` or whatever requires the least amount of casts.

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+          for (int i = 0; i < numInds; i++) {
+            fgl::Index index;
+            // get a value based on the data type indType, the array index i,
+            // and the byte stride bStride
+            switch (indType) {
+            case 5120:
+              index = (fgl::Index)indBuf[(bStride || 1) * i];
+              break;
+            case 5121:
+              index = (fgl::Index)(reinterpret_cast<unsigned char *>(
```
aardappel: you use C-style casts elsewhere..

dev/src/lobster/flat_gltf.h
```diff
@@ -0,0 +1,789 @@
...
+          int bStride = bv.stride;
+          // create an untyped char array from the index buffer offset, to be
+          // converted to a type
+          char *indBuf = &buffers[bv.buffer][bv.offset + a.offset];
+          for (int i = 0; i < numInds; i++) {
+            fgl::Index index;
+            // get a value based on the data type indType, the array index i,
+            // and the byte stride bStride
+            switch (indType) {
+            case 5120:
```
aardappel: I'm sure we can find the constant for this..

dev/src/lobster/flat_gltf.h
fluffrabbit: Infinite recursion. It is possible to create a circular parent-child-parent hierarchy in the glTF file, even though it would violate the spec. Node 1 "contains" Node 2, Node 2 "contains" Node 1, infinite loop. If someone wrote a glTF file by hand this could happen.

aardappel: In that case more elegant is to pass a `depth` param to `DrawTree` that you +1 on each rec call..
Even nicer to do this check at load time rather than at render time of course.

dev/src/gltfwrapper.cpp
fluffrabbit: `gl_load_scene_gltf` sorta sounds alright, but should it be something like `gltf_load_scene` and `gltf_render_scene` or `gl_load_scene_gltf` and `gl_render_scene_gltf` or `gl_load_gltf_scene` and `gl_render_gltf_scene`? I'd like to keep the word `scene` in there since as soon as the model is packed into the FlatBuffers format it's technically not a glTF-compliant file, but it's still a `scene`.

aardappel: Either way works for me.

dev/src/lobster/flat_gltf.h
fluffrabbit: I originally named `gltfwrapper.cpp` `glloadgltf.cpp`, but it also contains rendering code, so it does more than just loading. So there is the file naming question. There is also the question of how much code is reasonable to go in a single file. STB libraries go up to about 3000 lines, which is still maintainable, but source files in the Lobster codebase seem to be somewhat smaller. Organization is your choice since my private projects are structured completely differently anyways, though `gltfwrapper.cpp` containing all the code seems sensible based on your comments.

dev/src/lobster/flat_gltf.h
fluffrabbit: As an optional parameter, perhaps? I was under the impression that the path to the `default.materials` file was hard-coded.

dev/src/lobster/flat_gltf.h
fluffrabbit: Recursive lambdas unfortunately need an explicit `function` declaration, hence `#include <functional>`.

dev/src/lobster/flat_gltf.h
fluffrabbit: This gets into the bits and bytes that caused me the most problems when I first wrote the code. Leaving as is for the time being.

dev/src/lobster/flat_gltf.h
fluffrabbit: Returning Value( 0 or 1 ) through the nfr. Hopefully that works.

dev/src/lobster/flat_gltf.h
fluffrabbit: glTF does not use unsigned ints for counts, and that is the only cast in a comparison with count, so it's fine as an int. Anyhow, aren't the C++ people trying to phase out size_t?

dev/src/lobster/flat_gltf.h
fluffrabbit: One of those things I don't know how to fix.

dev/src/lobster/flat_gltf.h
fluffrabbit: Comments added.

dev/src/lobster/flat_gltf.h
fluffrabbit: Comments fixed. As for the blank lines, I have a hard time scanning code so I'm not sure which blank lines should stay and which blank lines should go.

lobster/schemas/gltf.fbs
fluffrabbit: `fgltf`, `flatg`, `fbtf`, dunno. I haven't given the `flatc`-compiled binary version much thought. One exports their `gltf` and copies it into the folder, and that's that. Binary versions just obscure it a little. The bulk of the data is in the binary chunk anyways, which currently has to be external until I implement `glb`, so maybe call full binary `glfb` and compiled JSON + external binary something else?

dev/src/lobster/flat_gltf.h
aardappel: I think having the loading & rendering code in the same file is fine.

I guess to me the cleanest (in the sense of fitting current Lobster) structure would be:

gltf_scene.h  // All the structs below that contain the scene loaded from a FlatBuffer. Depends as much as possible on glinterface.h. Has prototype decls for loading & rendering functions.

gltf.cpp  // Implements everything.

graphics.cpp  // (pre-existing) bindings for Lobster to load and render.

dev/src/lobster/flat_gltf.h
aardappel: It looks largely correct, if a bit complicated. Index buffers don't have strides, so you should be able to index with just `[i]`. What exactly is not working here?
Again, you need to find the exact opengl constants for these numbers. e.g. 5121 is 0x1401 which you can find in e.g. https://www.khronos.org/registry/OpenGL/api/GLES2/gl2.h (same for all versions)

lobster/schemas/gltf.fbs
aardappel: Not sure if there's a huge advantage to getting the external binaries internal.

dev/src/lobster/flat_gltf.h
aardappel: Yup, that's because it is needed in every single lobster program. Giving the user some control over files is generally good. Wouldn't make it optional.

fluffrabbit:
Not a performance advantage, especially with Emscripten where it's all one http request. It's just a convenience format.

lobster/schemas/gltf.fbs
fluffrabbit: No performance advantage. It's a convenience format. On the other hand, `gltfb` is an inconvenience format.

dev/src/lobster/flat_gltf.h
fluffrabbit: Okay, no idea what you mean by half of that; let me see if I understand.

* Rename `flat_gltf.h` to `gltf_scene.h` and `gltfwrapper.cpp` to `gltf.cpp`
* Offload some of the meat from `gltf_scene.h` to `gltf.cpp`
* Replace my `LoadTexture` function with the Lobster equivalent

dev/src/lobster/flat_gltf.h
fluffrabbit: Nope. It caches all the textures in video memory with associate filename strings so that all subsequent glTF loads recycle textures that are already loaded. It's the coolest way to deal with texture loading. 😎

dev/src/lobster/flat_gltf.h
fluffrabbit: `error: missing arg to builtin function: gltf_new_scene`

Huzzah, I just made it "better".

dev/src/lobster/flat_gltf.h
fluffrabbit: I don't think this goes to the index buffer, I think it goes to the vertex buffer, and all the byte alignment stuff has been too much of a challenge to give too much thought.

dev/src/lobster/flat_gltf.h
fluffrabbit: This is the kind of shit I don't like doing, which is why I think there should be [a plugin system](https://github.com/fluffrabbit/flat_gltf/blob/master/why_not_plugins.md).

dev/src/lobster/flat_gltf.h
fluffrabbit: I have no idea how to do that other than overwriting files and opening a new PR. I've had about enough of Git and GitHub.

dev/src/lobster/flat_gltf.h
fluffrabbit: Yeah, that `CreateTexture` function looks pretty familiar. ^.^ As nothing is rendering at this point and I don't want to screw it up worse, I'd like to know a bit more about how to use it.

The parameters are self-explanatory, but let me just confirm: `dim` should be `int2(width * height).data()` and `tf` is a bitmask thing where `TF_NONE` is retro/blocky and some combination of flags gives you the standard trilinear filtering.

`TF_BUFFER_HAS_MIPS` freaks me out a bit because OpenGL generates its own damn mips. Not my responsibility when using a texture loader, right? My `LoadTexture` function is heavily automated and checks for what byte alignment things should get, how mipmapping and filtering should be handled, etc. That way I just give it the image buffer and it fills in the blanks. Surely `CreateTexture` must also have an automatic mode.

lobster/schemas/gltf.fbs
aardappel: ?

dev/src/lobster/flat_gltf.h
aardappel: Yes, and move the actual bindings to graphics.cpp.
And make other code use lobster equivalents as well. Like the code generating vertex buffers etc.

dev/src/lobster/flat_gltf.h
aardappel: well, give it some more thought

dev/src/lobster/flat_gltf.h
aardappel: You can do the same with the result from `CreateTextureFromFile`.

And Lobster does this kind of caching in user code: https://github.com/aardappel/lobster/blob/master/lobster/modules/texture.lobster

dev/src/lobster/flat_gltf.h
aardappel: That would be nice, and is technically not hard: as you see the only function that really needs to be exposed to the Lobster core is the one that creates the bindings, and that could easily sit in a DLL. Then you could have a way to declare for a program which plugins it uses (since it is also needed at compile time).

It is a bit of work to structure things like that though, since Lobster builds on 6 platforms, not all of which have DLL mechanisms. So there would need to be a way to build it the old fashioned way as well. And I don't like having to depend on external repos for that.

Then there's the issue that some of these, like gltf, naturally share a lot with the rest of the "engine", so keeping that all separate would be a bit of duplication.

You're currently adding this to the core of the lobster engine, so it will have to conform to that engine. More work, but results in better code in the end. And I'll help maintain it from here on.

If you feel that having to bow to ugly conventions like snake_case makes it not worth contribution, you could have instead made a completely stand-alone gltf loading library in its own repo. Then you could have suggested it for use in Lobster by adding it to `dev/external/fluff_gltf` or whatever. The only things needing to go into the main code base would be the bindings. You could have kept your own coding style. On the downside, I would have still been picky on adopting it, as I only want well maintained high quality libraries in `external`, so more burden on you to prove that. And I wouldn't be maintaining any of it, so it had better be high quality at the moment I integrate it. Usually that means many github stars :P

dev/src/lobster/flat_gltf.h
aardappel: Sorry, I can't fix that for you.

And try `git clang-format`, it only formats the code in your PR.

dev/src/lobster/flat_gltf.h
aardappel: `dim` is a pointer because it could either supply an `int2` or `int3`.

There are many ways to create mips, and sometimes it makes sense to create your own. Some compressed texture formats even include them. Not sure why that is scary. It is an option, if you look at the code you'll see it calls `glGenerateMipmap` if that flag is not specified.

lobster/schemas/gltf.fbs
fluffrabbit: Facebook uses the GLB format because it's a quick export from a user's 3D program of choice, potentially bundling images as well as geometry data into one file. When I was in high school I would have loved it. `gltfb` is slightly to the side of that, but I will implement it when I get a chance when I figure it out.

dev/src/lobster/flat_gltf.h
fluffrabbit: Had I known what the `external` folder was for, I would have gone that route. Ah, well. If you want to absorb this thing into your body I should be grateful and help the process go smoothly.

As for shared libraries, the issues you mention are part of the whole DLL flavor. It wouldn't work on mobile and there would be lots of code redundancy, but those 10 MB DLLs would be someone else's problem, not yours. Current core libs are fine I guess, but I found your choice of IMGUI over other functionality a little strange from a game context. You can only add so many libraries before Lobster becomes bloated.

dev/src/lobster/flat_gltf.h
fluffrabbit: Well, it looks like I can just pop `CreateTextureFromFile` right into where I'm using the STB stuff and it should work. :)

dev/src/lobster/flat_gltf.h
fluffrabbit: Yeah, I always `glGenerateMipmap` when possible. The main caveats are image buffers and npot textures on GLES2, but this code goes back to the past, back to the `.dds` days. I remember those days...

dev/src/lobster/flat_gltf.h
fluffrabbit: It's using GL constants now. As for what is or isn't working, I don't know.

dev/src/lobster/flat_gltf.h
fluffrabbit: If I use Lobster's VAO loader, won't I be stuck with Lobster's current shaders? I mean, `phong_anim` is not the ideal vertex format for glTF/PBR, and I'm not sure if Lobster's VAO loader is generalized enough yet to handle new vertex formats.

dev/src/lobster/flat_gltf.h
fluffrabbit: 1git: 'clang-format' is not a git command. See 'git --help'.`

git 2.17.1

fluffrabbit: `git: 'clang-format' is not a git command. See 'git --help'.`

git 2.17.1

Rebased and all. Oh man, this site is a mess.

dev/src/lobster/flat_gltf.h
aardappel: Lobster's VAO code can deal with a variety of attributes, it is not fixed to one vertex format. They're also not tied to any particular shader, you can totally make your own.

Besides that, I see gltf and pbr as completely orthogonal issues, I don't see what they have to do with eachother. If rendering features are missing to support pbr, I am pretty sure we can add them.. either in this PR or maybe better later.

dev/src/lobster/flat_gltf.h
aardappel: Well, I expect you to know before this gets merged :)

dev/src/lobster/flat_gltf.h
aardappel: Again, just because you get an error doesn't mean the issue is resolved. I am sure you can find out how to either install clang-format for git, or just run clang-format manually on your own files.

fluffrabbit: `git clang-format-7` was needed on my machine. Resolved now.

dev/src/lobster/flat_gltf.h
fluffrabbit: Yeah, glTF and PBR meet at a perpendicular angle, much like vertex tangents and unlike vertex colors. But it's not like a model made with a PBR pipeline can't be kludged into Lobster's Phong pipeline, which is what I'm doing now. Once the damn thing renders (!) I will work on expanding Lobster's pipeline capabilities. Of course, I have avoided touching `graphics.cpp` up to this point, but you gave the ok so okay.

dev/src/lobster/flat_gltf.h
fluffrabbit: Everything (including the `reinterpret_cast`s) was working when I used my own renderer. When I began to switch to Lobster's renderer, everything stopped rendering. It's probably not a mismatch in vertex attributes as I think I fixed that (unless there's a typo). It could be a mismatch in uniforms, a matrix math issue, some weird data alignment corruption issue (as I suspect based on the blue flickering when you fly to the floor level), or another GL error (easily tested with glGetError*). So I'll have to work on that, but I doubt the problem is in this section of code.

dev/src/lobster/flat_gltf.h
fluffrabbit: Unresolving this as this is one of those places where things can get screwy.

aardappel:
Closing this for now. Once you feel all my past comments have been addressed, we can re-open with another review.

fluffrabbit:
damatrix
Most of these issues are addressed. However, nothing is showing up on the screen. I had [attached image] for a while but then it went away. Now I know it's a combination of matrix and shader issues.

aardappel:
Don't think anything in the rendering side of things has changed since you started?

fluffrabbit:
I see that a lot has changed, which forced me to reorganize. I've never been good at synchronized dancing, but to be clear, when I said that it stopped working, I meant that I broke it. That GIF is from just this morning after rebase. I broke it further when trying to conform it to the style of your code, like with currentshader->Activate() and other stuff from shader::Set(). :shrugs: I want it to work, but I understand my own code best of course.


-----Original Message-----

PR #40, opened 2019-04-12
Fix install target

jacereda:
Only tested on Darwin with Nixpkgs.

### DIFF
--- a/dev/CMakeLists.txt
+++ b/dev/CMakeLists.txt
@@ -12,20 +12,11 @@ set(SDL_STATIC ON CACHE INTERNAL "")
 set(VIDEO_VULKAN OFF CACHE INTERNAL "")
 add_subdirectory(external/SDL EXCLUDE_FROM_ALL)
 
-set(EXE_NAME lobster)
-
 file(GLOB LOBSTER_SRCS
     "src/*.h"
     "src/lobster/*.h"
     "src/*.cpp"
 )
-
-if(LOBSTER_TOCPP)
-  list(REMOVE_ITEM LOBSTER_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp")
-  list(APPEND LOBSTER_SRCS "compiled_lobster/src/compiled_lobster.cpp")
-  add_definitions(-DVM_COMPILED_CODE_MODE=1)
-  set(EXE_NAME compiled_lobster)
-endif()
 
 file(GLOB_RECURSE EXTERNAL_SRCS
     "include/Box2D/*.cpp"
@@ -104,27 +95,37 @@ if(WIN32)
   link_directories("$ENV{DXSDK_DIR}/Lib/$ENV{PROCESSOR_ARCHITECTURE}" "${CMAKE_SOURCE_DIR}/lib")
 endif()
 
-add_executable(${EXE_NAME} ${LOBSTER_SRCS} ${EXTERNAL_SRCS})
-
-target_link_libraries(${EXE_NAME}
-  ${SDL_LIBRARIES}
-  ${ADDITIONAL_LIBRARIES}
-  ${OPENGL_LIBRARIES})
-
 if(UNIX)
    include(GNUInstallDirs)
    set(BINDIR ${CMAKE_INSTALL_FULL_BINDIR})
    set(DATADIR ${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT_NAME})
    set(DOCDIR ${CMAKE_INSTALL_FULL_DOCDIR})
    set_source_files_properties(src/platform.cpp PROPERTIES COMPILE_FLAGS -DPLATFORM_DATADIR="\\\"${DATADIR}/\\\"")
+   if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+     set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+   endif()
 else()
    set(BINDIR ${CMAKE_INSTALL_PREFIX})
    set(DATADIR ${CMAKE_INSTALL_PREFIX})
    set(DOCDIR ${CMAKE_INSTALL_PREFIX}/docs)
-   if(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+   if(CMAKE_RUNTIME_OUTPUT_DIRECTORY)
      set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${BINDIR})
    endif()
 endif()
+
+set(EXE_NAME lobster)
+if(LOBSTER_TOCPP)
+  list(REMOVE_ITEM LOBSTER_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp")
+  list(APPEND LOBSTER_SRCS "compiled_lobster/src/compiled_lobster.cpp")
+  add_definitions(-DVM_COMPILED_CODE_MODE=1)
+  set(EXE_NAME compiled_lobster)
+endif()
+add_executable(${EXE_NAME} ${LOBSTER_SRCS} ${EXTERNAL_SRCS})
+target_link_libraries(${EXE_NAME}
+  ${SDL_LIBRARIES}
+  ${ADDITIONAL_LIBRARIES}
+  ${OPENGL_LIBRARIES})
+
 install(TARGETS ${EXE_NAME} DESTINATION ${BINDIR})
 install(DIRECTORY "${CMAKE_SOURCE_DIR}/../data" DESTINATION ${DATADIR})
 install(DIRECTORY "${CMAKE_SOURCE_DIR}/../modules" DESTINATION ${DATADIR})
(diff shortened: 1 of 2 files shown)


dev/src/platform.cpp
```diff
@@ -197,9 +197,7 @@ bool InitPlatform(const char *exefilepath, const char *auxfilepath, bool from_bu
             write_dir = exepath;
         }
         data_dirs.push_back(string(exepath));
-        #ifdef __linux__
-            data_dirs.push_back("/usr/share/lobster/");
-        #endif
+        data_dirs.push_back(DATADIR);
```
aardappel: Please surround with `#ifdef DATADIR`, this is built on many platforms with many build systems, not just CMake

dev/CMakeLists.txt
```diff
@@ -109,3 +109,15 @@ target_link_libraries(${EXE_NAME}
   ${SDL_LIBRARIES}
   ${ADDITIONAL_LIBRARIES}
   ${OPENGL_LIBRARIES})
+
+if(UNIX)
+   include(GNUInstallDirs)
+   set(BINDIR ${CMAKE_INSTALL_FULL_BINDIR})
+   set(DATADIR ${CMAKE_INSTALL_FULL_DATADIR})
```
aardappel: `DATADIR` is too to clash as a preprocessor symbol, maybe `PLATFORM_DATADIR` ?

dev/CMakeLists.txt
```diff
@@ -109,3 +109,15 @@ target_link_libraries(${EXE_NAME}
...
+else()
+   set(BINDIR .)
+   set(DATADIR .)
+endif()
+set_source_files_properties(src/platform.cpp PROPERTIES COMPILE_FLAGS -DDATADIR="\\\"${DATADIR}/lobster-data/\\\"")
```
aardappel: Why `lobster-data` instead of `lobster` ? The location already indicates it is data.

dev/CMakeLists.txt
```diff
@@ -109,3 +109,15 @@ target_link_libraries(${EXE_NAME}
   ${SDL_LIBRARIES}
   ${ADDITIONAL_LIBRARIES}
   ${OPENGL_LIBRARIES})
+
+if(UNIX)
+   include(GNUInstallDirs)
+   set(BINDIR ${CMAKE_INSTALL_FULL_BINDIR})
+   set(DATADIR ${CMAKE_INSTALL_FULL_DATADIR})
+else()
```
aardappel: remove `else` case? We don't want to add the current dir as `DATADIR`.

jacereda: I'm thinking of Windows here, so you can do something like this (untested)

```
cmake -DCMAKE_INSTALL_PREFIX=distribution .
make install
```

And end up with a `distribution` directory containing the required bits.

dev/CMakeLists.txt
jacereda: On non-unix platforms, the installed hierarchy will contain something like:
```
lobster
lobster-data/
```

I can't name both the directory and the executable `lobster`.

dev/CMakeLists.txt
jacereda: Right, done.

dev/src/platform.cpp
jacereda: Done.

dev/CMakeLists.txt
jacereda: Could be just `data` I guess, but then there will be a `data/data` directory.

dev/CMakeLists.txt
aardappel: On non-unix platforms, we're currently not using CMake, and we'd like to keep the directory structure as-is. E.g. both VS and Xcode produce an executable in the `lobster` dir. They don't need this additional data dir.

jacereda: Should I just remove the whole `if` and unconditionally use `GNUInstallDirs`?

dev/CMakeLists.txt
aardappel: We're not doing this on on non-unix platforms. Lets make sure these modifications you're making are only for platforms that desire to have data in `/usr/share` or similar. For others we're keeping existing layout.

On a side-note, I'd like to rename the `lobster` dir, since it is way too confusing in documentation and scripts. Not sure what to call it though.. any ideas?

dev/CMakeLists.txt
aardappel: No, again, don't want this to affect other platforms.

jacereda: Which non-unix platforms are built via CMake and what should a `make install` look like on those?

dev/CMakeLists.txt
jacereda: What about getting rid of the `lobster` directory and moving everything up one level?

dev/CMakeLists.txt
aardappel: None.
To be clear, we have these 6 platforms:

Windows: VS
OS X & iOS: XCode
Android: Android SDK/NDK
WebAssembly: Makefile
Linux: CMake

Theoretically other unices might build with the CMake file, but I don't think anyone has ever tried.

I may want to switch Windows to CMake at some point, but when I do, it shouldn't be trying to install files anywhere, just use the files in-place. So make install in system dir code only take effect on Linux for now.

dev/CMakeLists.txt
aardappel: I was going to move `docs` up, but it still makes sense to have the other files in one dir?

The original idea was the `lobster` dir is everything a user of the language needs, including the exe in that dir (which is Windows centric, sorry). Not sure if I want to dump all of that in the root.

Maybe you're right though.. will have to think about it.

aardappel: Well, you got your wish.. moved everything in `lobster/*` to the root. Easier this way.

dev/CMakeLists.txt
jacereda: What about this solution? I haven't tested, but on Unix installing to the sources dir seems to work. Files won't be copied (src and dst are the same), so only the lobster executable will be copied on `make install`.

aardappel: Again, on platforms that are non a Linux-alike (which is most of them), the existing behavior needs to be preserved. This changes the Windows exe for example to not be in `bin` anymore, but in some install dir? There is no "installing" on this platform. Please make your changes affect Linux only. I thought the whole point of this PR was just to allow us to do Linux packaging? So lets not use `UNIX` as it also includes all Apple platforms.

jacereda: - You said only Linux uses CMake.
- Even if you switched the Windows build to CMake, this behaviour is more reasonable. It should generate the lobster binary in the checkout directory. If you no longer want that behaviour, it should be easy to put it in a `bin` directory. 
- How can an `install` target for non-Unix be a bad thing? It should allow an easy isolation of required files for distribution (everything except the `dev` folder). If you just want to work in an install-less setup, just don't do `make install` and there you are.

jacereda:
Looks like the Linux CI is failing because there's a `lobster` directory present, could that be due to a previous build?

dev/CMakeLists.txt
aardappel: It currently does.. even so, OS-specific stuff should be in the proper conditionals. Even on Linux, it should end up in the `bin` dir during development, when not explicitly using `make install`. And there's nothing "more reasonable" given such a wide diversity of OS standards. The concept of "install" in this sense doesn't make sense for Windows (unless it generated an installer).

jacereda: It's already generating the unix executables under `bin`, check the last commit. And I have to disagree, `install` makes sense on Windows. Just tell it where to generate a distributable folder with `-DCMAKE_INSTALL_PREFIX=distributable`.

aardappel: Ok, I guess we disagree. If you insist, it can go in, I can always change it once I get to use it on Windows.

dev/CMakeLists.txt
```diff
@@ -4,26 +4,20 @@ project(Lobster)
...
+set(VIDEO_VULKAN OFF CACHE INTERNAL "")
+add_subdirectory(external/SDL EXCLUDE_FROM_ALL)
 
 file(GLOB LOBSTER_SRCS
     "src/*.h"
     "src/lobster/*.h"
     "src/*.cpp"
 )
 
-if(LOBSTER_TOCPP)
```
aardappel: Why was this removed?

jacereda: I moved it down so it's after the CMAKE_RUNTIME_OUTPUT_DIRECTORY setting, otherwise that setting won't take effect.

jacereda: TBH I didn't verify the `LOBSTER_TOCPP` case, but I assumed the `set(EXE_NAME` required `CMAKE_RUNTIME_OUTPUT_DIRECTORY`.

aardappel: Ah ok, didn't spot that.

aardappel:
Thanks!

aardappel:
So this did break the `bin` dir when using an "out of source" build, so had to revert it back to https://github.com/aardappel/lobster/commit/649d922d53541ab7b3880d19850414a3b0fa6dcc

Issue #42, opened 2019-04-14
Can't open file: stdtype.lobster


s-oram:
I'm attempting to create a simple Hello World on Windows. 

1. Lobster has been built. The `lobster.exe` was output to `D:\Dev\github\lobster\lobster\lobster.exe`.
2. Added `D:\Dev\github\lobster\lobster` to the Windows path. 
3. Created a new Lobster file `D:\Dev\lobster-hello-world\hello-world.lobster` with the contents:
```
print("Hello World")
```
4. Run `lobster hello-world.lobster` in the `D:\Dev\lobster-hello-world\` directory.
The result is `can't open file: stdtype.lobster`. 
I can see the `stdtype.lobster` file in the `\lobster\modules` directory. 

Is something else needed so that the Lobster compiler can find the files in the `lobster` directory? 

PS: I was able to compile `hello-world.lobster` by copying the `modules` directory to  `D:\Dev\lobster-hello-world\`, but I assume that is not the intended thing to do.

aardappel:
I've never tried putting the exe in the Windows path, so that may well not be working. I'll see if I can fix it.

Meanwhile, run it with its absolute (or even relative) path, that should work: `D:\Dev\github\lobster\lobster\lobster.exe D:\Dev\lobster-hello-world\hello-world.lobster` from anywhere.

aardappel:
Ok, found the problem, it is not using a more robust way to get the path.

s-oram:
Thanks Aardappel! I just tested with the recent changes and it now works! :)

PR #48, opened 2019-07-22
Fixes main dir issue on linux.

davidbruce:
Currently if you make a symlink to lobster in one of your path dirs, or just move lobster's folder into your path it will not be able to find your modules folder.  

This is the Linux equivalent of https://github.com/aardappel/lobster/issues/42


aardappel:
Thanks, that fits in perfectly :)

Issue #26, opened 2018-12-05
Mouse trapping on Linux


fluffrabbit:
Mouse trapping in a window is currently broken on some systems. The SDL best practice is: SDL_SetRelativeMouseMode

That works for me. I'm not sure how it fits into Lobster as the code comments imply some mouse-trap-switching with gl prefixed commands. The above SDL function allows me to Alt-Tab out of the window, but it still responds to mouse movement when out of focus, so checking for focus may still be desired.

aardappel:
Yup, SDL_SetRelativeMouseMode has had wonky behavior depending on the system for a while, and I even reported a bug to SDL about it that is still open after 3.5 years: https://bugzilla.libsdl.org/show_bug.cgi?id=2954. If you search, there's other people that have had trouble with it as well.

SDL is still the cross-platform layer (other options typically don't support ios/android well or at all), but there's tons of problems that don't get fixed. I've reported bugs that simply get closed un-fixed after years. And I really don't have time to go fix it for them.

Lobster indeed combines it with some other SDL calls: https://github.com/aardappel/lobster/blob/master/dev/src/sdlsystem.cpp#L613
This mode is intended for FPS style cameras, or other game modes where you don't want a system cursor.

fluffrabbit:
I need to work on my Github searching skills.

You disable it on Linux because it's broken on some Linuxes. Fair enough. On all platforms, there is this:
`if (fullscreen) SDL_SetWindowGrab(_sdl_window, SDL_TRUE);`

So it's impossible for mouse trapping to work in a window on Linux with the current code. The following works for me without issue. I wonder if the bogus values would be a big deal for Linux users with a broken SDL_SetRelativeMouseMode implementation when doing mouselook in a window, as the mouse hitting the edge of the screen also breaks it. Which is worse?

            if (fullscreen) SDL_SetWindowGrab(_sdl_window, SDL_TRUE);
            SDL_ShowCursor(0);
            #if defined(_WIN32) || defined(__APPLE__)
            // This is sometimes broken on Linux, giving bogus xrel/yrel in SDL_MOUSEMOVE
            SDL_SetRelativeMouseMode(SDL_TRUE);
            #else
            // Mouselook is broken without mouse trapping; hope for the best when windowed
            if (!fullscreen) SDL_SetRelativeMouseMode(SDL_TRUE);
            #endif
            clearfingers(false);

aardappel:
[...]

PR #49, opened 2019-07-28
Fix for mouse trapping issue on Linux. Issue #26

davidbruce:
Removes if def for SDL_SetRelativeMouseMode as issue is no longer reproducible with SDL 2.0.9.
I can not reproduce bad numbers coming from SDL_MOUSEMOVE xrel/yrel.   First person shooter camera from iqmtest works fine as does gl_mouse_pos(0) works as expected  with or without gl_grab.

### DIFF
--- a/dev/src/sdlsystem.cpp
+++ b/dev/src/sdlsystem.cpp
@@ -620,10 +620,7 @@ bool SDLCursor(bool on) {
         } else {
             if (fullscreen) SDL_SetWindowGrab(_sdl_window, SDL_TRUE);
             SDL_ShowCursor(0);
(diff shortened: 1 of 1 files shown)


aardappel:
Haven't confirmed this myself, but would make sense that it got fixed! Thanks :)

aardappel:
Looks like the version of SDL that's in Lobster's repo is 2.0.6, time to update I guess.

davidbruce:
It looks like 2.0.10 was just released on July 26th.

aardappel:
I just updated everything to 2.0.10

Issue #23, opened 2018-05-03
Discussion about the semantics of 'while'


shalabhc:
(Not sure if this is the appropriate forum for discussion but I didn't find a mailing list.)

I wanted to discuss the special evaluation semantics of expressions passed to `while` in the language. This actually confused me for a bit until I made my way to the end of the language reference (https://htmlpreview.github.io/?https://raw.githubusercontent.com/aardappel/lobster/master/lobster/docs/language_reference.html#built-in-functions). Before that I was looking for a way to pass unevaluated expressions around, but now I understand only calling `while` can pass an unevaluated expression (perhaps wrapping it in an anonymous function?) while *all* other calls pass the evaluated value.

Specifically, in the following code `expr` is passed unevaluated to `while`, but `expr2` and `expr3` are immediately evaluated and the value is then passed to the respective function:

```python
while expr:
   body

if expr2:
  body

for(expr3):
  body
```

From a design point of view, special casing `while` sticks out a bit. It breaks the expectation that expressions are evaluated immediately, but it does offer the convenient syntax. However it can still be confusing, e.g. I'm not sure how to pass an expression when calling while with the `while(...)` syntax, as opposed to the `while ...: ...` syntax.

I have a couple of questions:

1. Is there a way to define a function that accepts an unevaluated expression (i.e. can I implement `while` in lobster?) 
2. Are there an plans or ideas to 'sort out' the above exception around `while`?

I'm thinking if we have (1) above, then maybe `while` would not look like an outlier. OTOH, it means we would always have to look at the definition to know the call semantics.

aardappel:
This is totally appropriate place for discussion, thanks for the feedback!

The funny thing is that most programming languages (C/Java/Python..) have an exception for `while`: they all get an unevaluated expression, and it is the only expression to have this property (well, in C there's also `for` of course). To have more uniform syntax, `while` in C should read something like: `while { a < 10 } { a++; }` since C uses `{}` for unevaluated things. In Python, the condition should be a block also.

In Lobster it is actually less special case than you think, since that expression is actually represented in the compiler as a argument-less lambda, and treated as such during type checking and code generation. If the parser had a way for user defined functions to request this kind of parsing (e.g. a type `:exp`) then that would work without further changes to the language.

To currently write your own `while`:

    def mywhile(cond, do):
        while cond(): body()

    mywhile():
        a < 10:
    do:
        a++

[...]

PR #50, opened 2019-07-29
Issue #23 Adds lazy expression keyword. 

davidbruce:
Not sure if I did this correctly, but the testing I've done with it appears to be working. 

If the code change looks fine and you want to proceed with merging this please hold off.  I should probably make changes to the syntax files and add a small doc snippet.   

Side note this feature is fantastic and I wish I knew it was around sooner.

### DIFF
--- a/dev/src/lobster/parser.h
+++ b/dev/src/lobster/parser.h
@@ -527,6 +527,9 @@ struct Parser {
             case T_RESOURCE:  dest = type_resource;   lex.Next(); break;
             case T_LAZYEXP: {
                 lex.Next();
+                if (!funarg) {
+                  Error("lazy_expression cannot be used outside function signature");
+                }
                 funarg->flags = ArgFlags(funarg->flags | AF_EXPFUNVAL);
                 return -1;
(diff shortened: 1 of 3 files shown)


dev/src/lobster/parser.h
```diff
@@ -525,6 +525,11 @@ struct Parser {
             case T_STRTYPE:   dest = type_string;     lex.Next(); break;
             case T_COROUTINE: dest = type_coroutine;  lex.Next(); break;
             case T_RESOURCE:  dest = type_resource;   lex.Next(); break;
+            case T_LAZYEXP: {
+                lex.Next();
+                funarg->flags = ArgFlags(funarg->flags | AF_EXPFUNVAL);
```
aardappel: This should probably error if `funarg` is null, e.g. `struct A { x:lazy_expression }`.. otherwise LGTM!

davidbruce: I've added the error message.

aardappel:
Thanks!

PR #55, opened 2019-10-14
A couple of cosmetic improvements maid while orienting

jcorbin:
- align the signature of filter_indices with fiilter
- fix the gl_frame bool return docs


aardappel:
Thanks :)

PR #58, opened 2019-10-21
std.lobster: rework randomize to use fisher-yates

jcorbin:
This uses (one of the possible) interpretations of the [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher–Yates_shuffle), which is still fair, but does less array scanning than the push/remove prior.

I left in a `FIXME` note about how it'd be nice to have typical Pythonic (other languages like Go have also these days) multi-assign syntax.

### DIFF
--- a/modules/std.lobster
+++ b/modules/std.lobster
@@ -152,8 +152,11 @@ def insert_ordered(xs, x, lt):
     xs.push(x)
 
 def randomize(xs):
-    // Simple and effective, but remove makes it expensive.
-    for(xs.length) i: xs.push(xs.remove(rnd(xs.length - i)))
+    for(xs) x, i:
+        if i:
+            let j = rnd(i + 1)
+            xs[i] = xs[j]
+            xs[j] = x
     return xs
 
 def nest_if(c, nest, with):


modules/std.lobster
```diff
@@ -152,8 +152,14 @@ def insert_ordered(xs, x, lt):
     xs.push(x)
 
 def randomize(xs):
-    // Simple and effective, but remove makes it expensive.
-    for(xs.length) i: xs.push(xs.remove(rnd(xs.length - i)))
+    for(xs) x, i:
+        if i:
+            let j = rnd(i+1)
```
aardappel: Use formatting similar to the rest: `i + i` (spaces)

modules/std.lobster
```diff
@@ -152,8 +152,14 @@ def insert_ordered(xs, x, lt):
...
+            if j != i:
```
aardappel: given that these values are usually unequal, and that the code below is neutral to them being the same, it be cheaper to leave this `if` out?

aardappel:
Thanks!

PR #59, opened 2019-10-21
Add Will It Shuffle Sample

jcorbin:
Submitting my "hello world" program, an adaptation of Mike Bostock's [Will It Shuffle](https://bost.ocks.org/mike/shuffle/compare.html) comparison from the D3.js data viz world.

I used this program to both orient on the language, and to validate #58 .

Perhaps at least useful as a conversation piece, even if it's not a useful sample to merge since, at least for me, it points to potential language/stdlib improvements:
- clamp would be nice to use for numeric structs
- I've not fully climbed "Spline Maths Mountain", so I don't know yet if there's an easy answer to my Basis v Cardinal spline question
- exhibits my "singleton coroutine" situation, where in I wish I could write `let mycoro = coroutine def(): ...` to have an "anonymous singleton coroutine" (for shuffler and uier)
- a note about a missing strings stdlib, which I implement a single typical (join) function

### DIFF
--- a/dev/TODO.txt
+++ b/dev/TODO.txt
@@ -1,3 +1,15 @@
+- improved random number generation:
+  - swappable generators such as xorshift, pcg32/64, and even just a generic lcg
+  - a good lobster choice may be to abstract them as coroutines, or at least a
+    higher order `def generator(consume):` protocol
+  - especially so since insatiable random generators (to get multiple
+    independent streams) would be great in addition to / instead of a shared
+    global generator
+
+- a basis_spline equivalent of cardinal_spline
+
+- vectorized clamp function (and evaluate other adjacent functions while at it)
+
 - document tagged arg constructors.
 
 - See if a tail-call based interpreter would be a speed-up:
--- a/samples/will_it_shuffle.lobster
+++ b/samples/will_it_shuffle.lobster
@@ -1,0 +1,207 @@
+// Adaptation of https://bost.ocks.org/mike/shuffle/compare.html
+// Adapted by Joshua T Corbin <jcorbin@wunjo.org>
+//
+// TODO: would be nice to measure reshuffle time, in particular, adapt better
+// to expensive reshuffles that keep blowing the frame time budget; degrade
+// gracefully, and scale back the dataset size and/or round count
+
+import color
+import imgui
+import std
+import vec
+
+enum shuffle_mode:
+    shuffle_none = 0
+    shuffle_naive_swap
+    shuffle_push_remove
+    shuffle_fisher_yates
+
+var mode = shuffle_none
+
+def shuffle(array): switch mode:
+    case shuffle_none: array
+    case shuffle_naive_swap: naive_swap(array)
+    case shuffle_push_remove: push_remove(array)
+    case shuffle_fisher_yates: fisher_yates(array)
+
+def mode_name(): return switch mode:
+    case shuffle_none: "none"
+    case shuffle_naive_swap: "naive swap"
+    case shuffle_push_remove: "push remove"
+    case shuffle_fisher_yates: "fisher yates"
+    default: ""
+
+def naive_swap(xs):
+    // ala https://blog.codinghorror.com/the-danger-of-naivete/
+    for(xs) x, i:
+        let j = rnd(xs.length)
+        xs[i] = xs[j]
+        xs[j] = x
+
+def push_remove(xs):
+    for(xs.length) i:
+        xs.push(xs.remove(rnd(xs.length - i)))
+
+def fisher_yates(xs):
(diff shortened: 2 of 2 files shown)


samples/will_it_shuffle.lobster
```diff
@@ -0,0 +1,228 @@
+// adaptation of https://bost.ocks.org/mike/shuffle/compare.html
+//
```
aardappel: You can add yourself as the author, seeing as its one of the few files not by me.

samples/will_it_shuffle.lobster
```diff
@@ -0,0 +1,228 @@
...
+// TODO
+// - vectorize the b-spline
+// - more shufflers
+// - measure reshuffle time
+// - swappable RNG
```
aardappel: Feel free to add this to TODO.txt, or add a built-in function `set_rnd_generator` or whatever.

samples/will_it_shuffle.lobster
```diff
@@ -0,0 +1,228 @@
...
+        let j = rnd(xs.length)
+        xs[i] = xs[j]
+        xs[j] = x
+
+def push_remove(xs):
+    // std.lobster original
+    for(xs.length) i:
+        xs.push(xs.remove(rnd(xs.length - i)))
+
+def fisher_yates(xs):
```
aardappel: use vs in `std` if that gets merged

samples/will_it_shuffle.lobster
```diff
@@ -0,0 +1,228 @@
...
+    div( (_ >> 8)  & 0xFF, 255),
+    div(  _        & 0xFF, 255),
+    1,
+}
+
+def spline_value(values, t):
+    // uniforem b-spline interpolation adapted from d3-scale-chromactic
+
+    // TODO can we translate from this Basis Spline (B-spline) to a cardinal
+    // spline? Or should lobster also provide a builtin for B-splines?
```
aardappel: Yup, feel free to add other spline functions, similar to the code for cardinal spline. That is nice to have, and generally good to have very re-usable core code in C++.

samples/will_it_shuffle.lobster
```diff
@@ -0,0 +1,228 @@
...
+
+    t = clamp(t, 0, 1)
+    let n = values.length - 1
+    let i = clamp(floor(t * n), 0, n-1)
```
aardappel: And yes, if there's no vector clamp, there should be. Vector all the things!

samples/will_it_shuffle.lobster
```diff
@@ -0,0 +1,228 @@
...
+
+        if shuffling or changed or last_mode != mode:
+            reshuffle()
+            last_mode = mode
+        next(nil)
+    return nil
+let shuffler = coroutine shuffle_routine()
+
+def viz():
+    let status = join(" ", [
```
aardappel: Doesn't this do pretty much exactly what `join` does?

https://github.com/aardappel/lobster/blob/dffd8a8944cbe5d81789b22926f725b013183cd7/dev/src/builtins.cpp#L504-L515

jcorbin: Oops, had missed that one.

samples/will_it_shuffle.lobster
jcorbin: Lofted most out into `dev/TODO.txt`, and expanded on my own internal musings on where the PRNG stdlib should head...

samples/will_it_shuffle.lobster
jcorbin: Simplified to match, but leaving both dual implemented here for clarity of the demo fwiw

samples/will_it_shuffle.lobster
jcorbin: Neat, will look into doing so after this PR; I think I'm going to actually try writing something more like a game first fwiw :-)

samples/will_it_shuffle.lobster
jcorbin: Yup, left it as a TODO for now, rather than keep sprawling out in this PR :-)

dev/TODO.txt
```diff
@@ -1,3 +1,15 @@
+- improved random number generation:
+  - swappable generators such as xorshift, pcg32/64, and even just a generic lcg
+  - a good lobster choice may be to abstract them as coroutines, or at least a
+    higher order `def generator(consume):` protocol
```
aardappel: there's currently no way to create a "builtin co-routine" (one written in C++). I'd suggest keeping this simple: just a function that can select which type of rng is currently active.

aardappel:
Thanks! Future improvements can be in followups.

PR #62, opened 2019-11-02
Sorting Things Out

jcorbin:
Example results on my machine before the change to use insertion sort:

```
N:10    ops:6457        s/op:0.000154881044
N:100   ops:381 s/op:0.002627428871
N:200   ops:165 s/op:0.006089478788
N:400   ops:74  s/op:0.013641381081
N:1000  ops:25  s/op:0.04020962
```

However what's really prompting me to toss this into an early draft PR, is this mysterious error:
```
std.lobster(129): VM error: VM internal assertion failure: C:\Users\jcorb\lobster\dev\src\vm.cpp: 402: sp == stf.spstart
   stack: 0xa
in function: rec -> std.lobster(129)
   l = 10
   e = 10
   s = 0
in function: qsort_in_place -> std.lobster(146)
   lt = <FUNCTION>
   xs = [8, 6, 7, 9, 5, 1, 4, 2, 0, 3]
in function: function114body -> sorttest.lobster(29)
   ys = [8, 6, 7, 9, 5, 1, 4, 2, 0, 3]
in function: function111fun -> sorttest.lobster(18)
   t0 = 0.0360891
in function: map -> std.lobster(6)
   i = 0
   x = 0
   r = []
   fun = <FUNCTION>
   xs = 1
in function: timeit -> sorttest.lobster(16)
   n = 1
   times = []
   body = <FUNCTION>
   limit = 100000
   budget = 1.0
in function: __top_level_expression -> sorttest.lobster(26)
   xs = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
   N = 10
```

Also do note the typed-0 weirdness regarding `sum`.

### DIFF
--- a/tests/sorttest.lobster
+++ b/tests/sorttest.lobster
@@ -1,0 +1,33 @@
+import std
+
+// FIXME why is the `0` literal typed
+def sumf(xs): return fold(xs, 0.0): _x + _y
+
+def timeit(budget, limit, body):
+    var times = []::float
+    while times.length() < limit and times.sumf() < budget:
+        let n = if times.length() == 0:
+            1
+        else:
+            let total = times.sumf()
+            let avg = total / times.length()
+            let rem = budget - total
+            min(limit, max(1, int(rem / 2 / avg)))
+        let more = map(n):
+            let t0 = seconds_elapsed()
+            body()
+            let t1 = seconds_elapsed()
+            t1 - t0
+        times = append(times, more)
+    return times from timeit
+
+for([10, 100, 200, 400, 1000]) N:
+    var xs = map(N): _
+    let times = timeit(1.0, 100000):
+        var ys = copy(xs)
+        ys.randomize()
+        ys.qsort_in_place(): _a < _b
+        assert equal(xs, ys)
+    let total = times.sumf()
+    let avg = total / times.length()
+    print "N:" + N + "\tops:" + times.length() + "\ts/op:" + avg


modules/std.lobster
```diff
@@ -119,21 +119,30 @@ def qsort(xs, lt):
...
-                    sp++
-                else:
-                    xs[sp] = xs[--ep]
-                    xs[ep] = c
-            xs[--sp] = pivot
-            rec(s, sp)
-            rec(ep, e)
+
+        switch l:
+            case 0, 1:
```
aardappel: why this over `if l <= 1:` ?

modules/std.lobster
```diff
@@ -119,21 +119,30 @@ def qsort(xs, lt):
...
+                return
+
+        if l <= 32:
+            insertion_sort(slice(xs, s, l), lt)
```
aardappel: I'm sure this might be quicker, but it says `qsort` on the box, so it should do qsort :) Lets not mix sorting algos. Besides being a stdlib, its also an example, so certain level of elegance is good. And small is good too, seeing as how these functions typically all get specialized.

modules/std.lobster
```diff
@@ -119,21 +119,30 @@ def qsort(xs, lt):
...
-                if lt(c, pivot):
-                    xs[sp - 1] = xs[sp]
-                    sp++
-                else:
-                    xs[sp] = xs[--ep]
-                    xs[ep] = c
-            xs[--sp] = pivot
-            rec(s, sp)
-            rec(ep, e)
+
```
aardappel: style: no need for these empty lines.

tests/sorttest.lobster
```diff
@@ -0,0 +1,33 @@
+import std
+
+// FIXME why is the `0` literal typed
```
aardappel: What is the problem? `0` is an `int` in most statically typed languages.

jcorbin: Well the problem is typing: you can't use `sum(xs)` on an `xs` of type `[float]` because the `0` initial accumulator to `fold` is an int.

Hence my need to define `sumf` here which starts a `fold` out with `0.0` instead.

Reasons like this are why, e.g. in Go, literals and even numeric constants (as in `const foo = 42`) untyped; if instead you want a typed constant, you do `const foo int = 42` or `const foo float64 = 42`.

modules/std.lobster
jcorbin: Right, this PR is a draft to demonstrate a crash, this part isn't actually a serious "I think we should actually change it" yet; the crash is the only thing here, and not even really "for review" as this is a draft PR ;-)

modules/std.lobster
jcorbin: Agree, again this is just me experimenting, and shared only to demonstrate a crash that I'm having trouble sorting out (pun very much intended)

jcorbin: Although, it's very rare in practice for anything calling itself "quicksort" to be "quicksort all the way down"; that's usually only in teaching example code ;-)

tests/sorttest.lobster
aardappel: While I agree that it is inconvenient here, typically it is the right thing to do, and what most static languages do. Go is again the odd one out here (as it is with a lot of things). Being able to get the int or float version of a function by just using a constant is usually very convenient.

None of these languages have the "automatic genericity" of Lobster though, which makes this a special case. Really the best way to solve this would be to allow explicit type parameters, e.g. `def sumf<T>(xs:[T]): return fold(xs, T(0)): _x + _y` or similar.

For now, my solution would be to add a starting value to `sum` rather than define a second version.

Note `weighted_pick` right above there in `std.lobster` has the exact same problem. I guess its time for explicit template parameters :)

aardappel: https://github.com/aardappel/lobster/commit/a9011897b9870c42e468ae75b44b6b50c9f7c3ab

aardappel:
Ok, this fixes the assertion: https://github.com/aardappel/lobster/commit/b6662d98e76e1e7ae4941b8c83f6dd605acd3d86
It was due to your interesting `switch` :)

PR #64, opened 2019-11-04
Add basic image example

kriskowal:
This adds a sample that renders the lobster image, filling the window, as a small stepping-stone for beginners.


aardappel:
Nice! thanks :)

PR #65, opened 2019-11-04
Add centered image grid example

kriskowal:
This example illustrates how to translate and scale a unit square to fill the center of the window, then fills that unit square with lobsters in five ranks and files, using the forxy 2d iterator defined in modules/vec.lobster.

### DIFF
--- a/samples/imagegrid.lobster
+++ b/samples/imagegrid.lobster
@@ -1,0 +1,32 @@
+// This example illustrates how to translate and scale a unit square to fill
+// the center of the window, then fills that unit square with lobsters in five
+// ranks and files, using the forxy 2d iterator defined in modules/vec.lobster.
+
+import gui
+import texture
+
+fatal(gl_window("Lobsters Rank and Vile", 512, 512, 1))
+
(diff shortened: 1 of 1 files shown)


aardappel:
Thanks! This example is pretty similar to the previous one though, not sure if showing how to put images in a grid is worth another example. Also I'd inline `center_unit_square` and use a comment instead, to make it easier to follow. Maybe the two examples can be merged such that the useful learnings can all be in one?

kriskowal:
Yeah, I’m tracing my steps since it naturally progresses as another graduated tutorial. I’m tempted on one hand to combine these into a single example, but that risks obfuscating the primary insight.

Instead of landing this, I could queue up a sequence of elaborations for a sprites and tiles tutorial as I pick up a concept at a time, or drop it if it turns into a dead end.

aardappel:
Or if its really a tutorial, add something similar to http://aardappel.github.io/lobster/shooter_tutorial.html :)

kriskowal:
Indeed. I’ll see where this takes me.

Issue #67, opened 2019-11-09
Feature request: choose random element


kriskowal:
This is a request for design feedback before submitting a change.

A “choose one from vector” method would be handy in the standard library.

```
def foo(xs): xs[rnd(length(xs))]
```

I’ll follow up with some messages we can 👍 or 👎 for specifics details.

kriskowal:
Option: name the method `pick`

kriskowal:
Option: name the method `choose`

kriskowal:
Option: overload `rnd`

kriskowal:
Option: export from standard library

kriskowal:
Seems like there will be an opportunity to also provide another overload like `foo([1, 2, 3], 2)` might yield `[1, 2]`, `[2, 3]`, or `[1, 3]`.

aardappel:
I'd suggest `rnd_pick` since we already have a family of `rnd` functions.

We could add it to `std`, but its so common I can also just add it to the builtins.

aardappel:
Decided in the end to just add them to `std`, since I also had use for a 2D version in my private code: https://github.com/aardappel/lobster/commit/290203359dc16f60b96c709587185075a5cef110

PR #69, opened 2019-11-09
Add Langusta de Langton demo

kriskowal:
This is a demonstration of Langton’s Ant https://en.wikipedia.org/wiki/Langton%27s_ant

With one lobster, this is a faithful representation of Langton’s Ant, implemented as a cellular automaton instead of a single-agent automaton.  However, there are additional rules that dictate the behavior of collisions.  Each lobster has a unique priority.  Each lobster makes a bid to follow the course set out for them by the langton rule number beneath them.  For each tile, the neighboring lobster with the highest priority that has bid to move onto that tile advances.  All lobsters that fail their bid turn in place instead.


samples/demos/langtons.lobster
```diff
@@ -0,0 +1,311 @@
...
+// langton rule number beneath them.
+// For each tile, the neighboring lobster with the highest priority that has
+// bid to move onto that tile advances.
+// All lobsters that fail their bid turn in place instead.
+
+import gui
+import texture
+
+def pick(list):
+    return list[rnd(length(list))]
```
kriskowal: To be removed after #67.

samples/demos/langtons.lobster
```diff
@@ -0,0 +1,311 @@
...
+// to the next.
+def transit(len, dst, src, transitions):
+    for(len) i: dst[i] = src[transitions[i]]
+
+// move copies one array to another, given the length of both.
+// This is necessary at the beginning of each tick, to overwrite columns of
+// data from the previous generation.
+def move(len, dst, src):
+    for(len) i:
+        dst[i] = src[i]
```
kriskowal: Would be neat to have a memcopy builtin for this case. Maybe overload copy, for the no-allocate case. I assume `min(length(src), length(dst))` would be cheap, but I have that figure laying around in this case.

aardappel: Yup that be nice to have

aardappel:
Thanks!

PR #70, opened 2019-11-10
Add cube rendering example

kriskowal:
This example renders a cube with a different colorèd lobster on each face.  Moving the mouse or WASD rotates the viewport around the cube like turning a globe in front of the camera.

The example illustrates creating a simple mesh and rendering a texture on the mesh in various orientations.  The example showcases the render_to_texture utility, which allows us to render a 2D drawing on the surfaces of a 3D object, using a frame buffer.

### DIFF
--- a/samples/cube.lobster
+++ b/samples/cube.lobster
@@ -1,0 +1,150 @@
+// This example renders a cube with a different colorèd number on each face.
+// Moving the mouse or WASD rotates the viewport around the cube like turning a
+// globe in front of the camera.
+//
+// The example illustrates creating a simple mesh and rendering a texture on
+// the mesh in various orientations.
+// The example showcases the render_to_texture utility, which allows us to
+// render a 2D drawing on the surfaces of a 3D object, using a frame buffer.
+// The example also demostrates using gl_new_mesh, which allows us to orient
+// the texture on each face of the cube, since the simpler gl_new_poly is not
+// quite sufficient for this purpose.
+//
+// To generate the mesh for each face, cube_face_meshes uses the three least
+// significant bits of ASCII characters to represent whether the edge is on the
+// near or far side of the cube along each axis.
+//
+//  -----ZYX CHR corner
+//  00110000 '0' origin
+//  00110001 '1' x
+//  00110010 '2' y
+//  00110011 '3' xy
+//  00110100 '4' z
+//  00110101 '5' xz
+//  00110110 '6' yz
+//  00110111 '7' zyz
+//
+// Each face of the cube contains four of the cube's vertices.
+// The normal vector for the face of the cube must face outward to be opaque to
+// an outside observer, so the vertices are listed counter-clockwise.
+// In the following illustration, the interior faces are inverted, so the
+// vertices appear in clockwise order from our perspective.
+//
+// The first index must be the top-right corner of the texture.
+// The textures are arranged such that the textures are upright around the
+// equator and the poles are connected top to bottom with their nearest
+// neighbor.
+// Rotating the vertex strings rotates the corresponding texture orientation.
+//
+// The faces are numbered according to the conventions of right-handed dice.
+// All faces in opposition have the same sum.
+// Numbers read counter-clockwise about the origin and its opposite vertex.
+//
+//                                 inverted clockwise
+//  Z       4---5             .---.   .---.
+//   \      |\  |\    1540->2 |\ 2 \  |   |\  6<-5464
+//    0--X  | 0---1           | .---. | 6 | .
+//    |     | | | |   0462->3 |3|   | |   |4| 4<-5137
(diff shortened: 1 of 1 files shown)


samples/cube.lobster
```diff
@@ -0,0 +1,111 @@
...
+let detail = 256
+let face_tex = map([
+    color_purple,
+    color_olive,
+    color_teal,
+    color_dark_red,
+    color_dark_blue,
+    color_dark_green,
+]) color:
+    render_to_texture(nil, xy_1i * detail, false, nil, 0):
```
aardappel: while it is nice to demo this functionality, this is a really heavy-weight way to do coloring.. instead you can simply call `gl_color` in the `for(6)` loop below..

samples/cube.lobster
```diff
@@ -0,0 +1,111 @@
...
+        gl_set_primitive_texture(0, lobster_tex)
+        // The texture colors are multiplied by the color in context.
+        // Since our texture is white on black, we can change the white to a
+        // unique color for each face of the world.
+        gl_color(color)
+        gl_rect(xy_1 * detail, 0)
+
+def transform_to_face(i, f):
+    gl_translate(xyz_0):
+        if i == 0: // -x
```
aardappel: switch?

samples/cube.lobster
```diff
@@ -0,0 +1,111 @@
...
+            gl_translate(xyz_x / -2)
+            gl_rotate_z(-xy_y)
+            gl_rotate_x(-xy_y)
+            gl_translate((xyz_x + xyz_y) / -2)
+        else: if i == 1: // -y
```
aardappel: wonder if theres a more elegant way to do this. in particular, the `lobstercraft.lobster` achieves this by simply generating 6 faces with different params.

samples/cube.lobster
```diff
@@ -0,0 +1,111 @@
...
+        else: if i == 5: // +x
+            gl_translate(xyz_x / 2)
+            gl_rotate_z(xy_y)
+            gl_rotate_x(-xy_y)
+            gl_translate((xyz_x + xyz_y) / -2)
+        f()
+
+// This demo is able to use camera_FPS_view but uses a different control model
+// to move the camera about the origin at a fixed “elevation”.
+def camera_GPS_update(upkey, leftkey, downkey, rightkey, elevation:float, mousesens:float, keyspeed:float):
```
aardappel: there is some of this functionality already in `3dutil.lobster` that may be reusable?

samples/cube.lobster
kriskowal: I think to rationalize the demo, I’ll render the face number instead of the lobster, and also move the color out. I didn’t realize the color mask would thread through textures on meshes!

samples/cube.lobster
kriskowal: I’d like to use an array of transform matrices. That would eliminate the function entirely.

I saw the trick in lobstercraft too. Now that I better understand how it works, I might have more luck. It would be interesting to see how it treats the orientation of the textures. One of the motivations of this approach was versatility. From this jump-off, I can render a selection box on the surface on top of the cube, for example. Or, dynamically render the surface texture on the frame buffer. One way I can run with this is to do a CA demo on the cube surface.

samples/cube.lobster
kriskowal: This is pretty different from camera_FPS_update, since it keeps the camera pointed at the center of the model at a fixed distance.

I did not see another function that duplicates this behavior, but I’ll look again.

samples/cube.lobster
kriskowal: Pushed this, so it renders numbers instead of lobsters.

aardappel:
I don't mind merging this as-is, if you're done with it, as it is useful to demonstrate the many rendering calls it demonstrates. That said, it remains a complex way to render a cube, with the big switch of transforms and the render to texture not things I would generally recommend people to use like this.

samples/cube.lobster
aardappel: Ok, fair enough.

samples/cube.lobster
aardappel: To create such matrices, you'd still need functions similar to the ones you're calling, so it wouldn't do much for code complexity. Generally though, rendering something like this with transforms doesn't seem like the way to go.

kriskowal:
I’ll look into using the approach from lobstercraft and see if I have the
same control over texture orientation. Changes planned.

samples/cube.lobster
kriskowal: I’ve adapted the solution from LobsterCraft™, and basically wrote my dissertation on reverse engineering it. Hope that helps.

kriskowal:
Back to you in the studio, @aardappel. This now uses the connect-the-vertices approach.

aardappel:
Awesome.. sweet ascii diagrams! :)

PR #72, opened 2019-11-15
Exploit vector remainder in Langton’s Lobsters

kriskowal:
This change takes advantage of the new remainder operator’s behavior on vectors, replacing a one-off function in the Langton’s Lobsters demo.


aardappel:
nice!

PR #74, opened 2019-11-17
std.lobster: add partition_indices

jcorbin:
Add an index-oriented version of `partition(xs, fun)`, similar to how we already  have `filter(xs, fun)` and `filter_indices(xs, fun)`.


aardappel:
Makes sense, thanks!

PR #82, opened 2019-11-25
Move texture computation out of util3d

jcorbin:
Moves all of the texture processing utilities out of `util3d` into `texture`, separating them from the 3D/FPS camera affair there-in.

Original impetus was that the camera variables were "polluting" my `im_show_vars`  , despite me not needing a 3d camera. I'd worked around by copying / adapting the function in the short term.

But on balance, this seems to be the better home for these functions.

Maybe eventually "util3d" should be pivoted into a camera module?


modules/texture.lobster
```diff
@@ -45,3 +45,39 @@ def loadtexturecached(name, tf:texture_format):
             return from program
 
 def loadtexturecached(name): return loadtexturecached(name, texture_format_none)  // FIXME: default args.
+
+// Pass nil for tex to have it created. Recreates tex if the wrong size. Renders f. Returns tex.
+// Depth texture automatically created if depth is true and depthtex is nil.
+def render_to_texture(tex, size, depth, depthtex, texture_format, f):
+    if !tex || gl_texture_size(tex) != size:
+        tex = gl_create_blank_texture(size, color_black, texture_format)
```
jcorbin: Later realized: we should change the initial color her to `color_clear`:
- it's not possible to `gl_clear(color_clear) // clear to transparent`
- but you can clear to black

aardappel:
Yup that makes a lot more sense!

modules/texture.lobster
aardappel: it's kinda best the user calls `gl_clear` themselves, since it is expensive, and don't want to do it twice.. in fact sometimes its ok to not do it at all.

aardappel:
Missing: https://github.com/aardappel/lobster/commit/0428a3aa709900c37a633962cc1b4135124096fe

PR #83, opened 2019-11-28
texture: use clear color in render_to_texture

jcorbin:
This makes it easier to generate partially transparent textures, e.g. for a sprite sheet that aren't all solid squares.


aardappel:
Thanks!

PR #84, opened 2019-11-28
std: provide index under reverse() ; add reverse_map

jcorbin:
Provide the element index for symmetry with `for()`

### DIFF
--- a/modules/std.lobster
+++ b/modules/std.lobster
@@ -106,7 +106,14 @@ def product(xs): return fold(xs, 1): _x * _y
 
 def zip(xs, ys): return map xs.length: [ xs[_], ys[_] ]
 
-def reverse(xs, fun): for(xs.length) i: fun(xs[xs.length - i - 1])
+def reverse(xs, fun): for(xs.length) i:
+    let j = xs.length - i - 1
+    fun(xs[j], j)
+
+def reverse_map(xs, fun): return map(xs.length) i:
+    let j = xs.length - i - 1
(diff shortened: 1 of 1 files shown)


modules/std.lobster
```diff
@@ -106,8 +106,8 @@ def product(xs): return fold(xs, 1): _x * _y
 
 def zip(xs, ys): return map xs.length: [ xs[_], ys[_] ]
 
-def reverse(xs, fun): for(xs.length) i: fun(xs[xs.length - i - 1])
-def reverse_list(xs): return map(xs.length) i: xs[xs.length - i - 1]
+def reverse(xs, fun):        for(xs.length) i: fun(xs[xs.length - i - 1], xs.length - i - 1)
```
aardappel: Maybe nice to put `i = xs.length - i - 1` so to only compute that once? Lobster doesn't have a particularly advanced optimizer yet, so particularly for functions in `std` its worth to write the most optimal code manually :)

Also, there seem to be extra spaces in random places?

jcorbin: Done; I for one prefer to dedupe my own sub-expressions, rather than posit the mythical "sufficiently advanced compiler"; I only hadn't done here for minimal change previously ;-)

Also introduced `reverse_map` on argument of "obviousness" and "completeness".

Walked back the, now obsolete, white space changes for vertical alignment (in the prior regime of compact one-liners).

aardappel:
Thanks!

PR #87, opened 2019-12-28
Get commit hash and date from git in CMake, use in main.cpp

### LINKED ISSUE #86: Build version number into executable
I don't see a way to confirm that I'm running a specific version of Lobster. When I build locally, and run without args I [get the build date](https://github.com/aardappel/lobster/blob/21117d53a5ccedfdb03da732d6d881df1ec47fe9/dev/src/main.cpp#L169), which is certainly better than nothing, but doesn't really tell me the state of the sources when I built.

One solution is to automate incorporating the git tag and hash (and branch if being ambitious) into the binary, and have Lobster print that on the console along with the build date. We have a script to do that at work; I can't share it, but Googling around I found a couple similar things

https://www.mattkeeter.com/blog/2018-01-06-versioning/

or

https://github.com/bufferoverflow/proxyme/blob/master/CMakeLists.txt#L5

```
lobster e% git rev-parse --short HEAD 
21117d5
lobster e% git diff-index --name-only HEAD
lobster e% git name-rev --tags --name-only 21117d5
undefined
```

These are stuffed into a file with CMake's `configure_file`

dcurrie:
This commit fixes #86 for CMake by getting commit hash in CMake and providing to `main.cpp` via a preprocessor macro `GIT_COMMIT_INFOSTR` 

It's does not yet address the issue in the other build systems. I looked at XCode, and I'm not sure if the preprocessor macro mechanism will work without a non-versioned include or .xcconfig file.

### DIFF
--- a/dev/CMakeLists.txt
+++ b/dev/CMakeLists.txt
@@ -9,6 +9,15 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX ${CMAKE_SOURCE_DIR}/..)
 endif()
 
+# git commit date and commit hash into a definition
+execute_process(
+  COMMAND git log -1 --pretty="%cI|%h"
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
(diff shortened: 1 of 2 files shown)


aardappel:
Cool, thanks!

PR #90, opened 2020-01-03
Add hash overloads for int and float

### LINKED ISSUE #89: hash for non-reference values
The (auto generated) builtin functions reference says:
```
hash(x) -> int | hashes any value into an int
```
but this:
```
    assert hash(1) is int
```
gives:
```
error: "hash" (x argument) requires type: reference type, got: int
```
and the same problem for `float` args.

I work around it for now with 
```
if key is int: key else: if key is float: hash(string(key)) else: hash(key)
```

dcurrie:
Fixes #89

### DIFF
--- a/dev/src/builtins.cpp
+++ b/dev/src/builtins.cpp
@@ -1103,16 +1103,28 @@ nfr("active", "coroutine", "C", "B",
         return Value(active);
     });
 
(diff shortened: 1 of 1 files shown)


aardappel:
Thanks!

PR #91, opened 2020-01-21
Fix typo on font path

fgaz:
Just a typo I noticed while trying to run the samples


samples/guitest.lobster
```diff
@@ -1,7 +1,7 @@
 import gui
 
 fatal(gl_window("gui test", 960, 640))
-check(gl_set_font_name("data/fonts/US101/US101.ttf"), "can\'t load font!")
+check(gl_set_font_name("data/fonts/US101/US101.TTF"), "can\'t load font!")
```
jcorbin: This likely is a problem for Linux users, while Windows and Mac usually do case conversions in their fs calls.

Fwiw, confirming that the git tree object does agree that it should be all caps:


```
$ git rev-parse master
3c046d508c5a6423c040e5dd9aa6121965573093

$ git rev-parse master:data/fonts/US101
5c971b633a277217e0007b17193740238d505c7c

$ git ls-tree master:data/fonts/US101
100644 blob 9850a0629065a95838ae39732990db7567cdc119    US101.TTF
100644 blob 4f9d8dc463a2e1fa60f04d01c88980dfb81f7b41    US101.txt
```

jcorbin: Zooming out, question for @aardappel, another direction this sort of thing could go would be to say "Lobster should standardize such cross-platform case coalescing behavior"; thought?

aardappel: @jcorbin I'm not even sure what the correct solution here would be. Windows programmer writes a file name with wrong case, it fails on Linux. What does Lobster do on Linux? Try all permutations of file name cases? Scan dir, lowercase all the file names, and find it that way?

But on linux, having 2 filenames with differing case is legal. In fact, a Linux programmer expects it to fail, and would not be happy with case errors going unnoticed.

Probably the most elegant error would be to make Windows conform to Linux: find the file, but if case doesn't match, fail as if the file wasn't there :) But this will confuse Windows users also.

aardappel:
Thanks! For this particular case, I'd say just renaming the file would be even better, seeing as how all other files in `data/fonts` end in `.ttf`

samples/guitest.lobster
aardappel: Maybe a warning on Windows. There's no good reason why Windows user should ever address a file under the wrong case. I'm guessing case insensitivity is a useful feature when typing filenames manually in the console, but not in source code.

fgaz: +1 for current behaviour. Let linux be linux and windows be windows :)

fgaz:
Sure. I'll push a rename tomorrow

fgaz:
Updated

aardappel:
Thanks!

PR #97, opened 2020-06-08
Allow `is` to be used with unspecialized generic type

arvyy:


dev/src/lobster/typecheck.h
```diff
@@ -313,6 +323,13 @@ struct TypeChecker {
         return a == b;  // Not inlined for documentation purposes.
     }
 
+	bool TypeOfUndefinedGenericSpec(TypeRef t) {
+		if (t->udt == nullptr) return false;
```
aardappel, line 327: This is not a safe check in a union, you likely want to call `IsUDT` here

dev/src/lobster/typecheck.h
```diff
@@ -3149,10 +3168,10 @@ bool Not::ConstVal(TypeChecker &tc, Value &val) const {
 }
 
 bool IsType::ConstVal(TypeChecker &tc, Value &val) const {
-    if (child->exptype == resolvedtype || resolvedtype->t == V_ANY) {
-        val = Value(true);
-        return true;
-    }
+	if (tc.ConvertsToStatically(resolvedtype, child->exptype)) {
```
aardappel, line 3171: Make sure you use the existing indentation style, here and elsewhere.

dev/src/lobster/parser.h
```diff
@@ -584,6 +584,7 @@ struct Parser {
             case T_STRTYPE:   dest = type_string;     lex.Next(); break;
             case T_COROUTINE: dest = type_coroutine;  lex.Next(); break;
             case T_RESOURCE:  dest = type_resource;   lex.Next(); break;
+			case T_ANYTYPE:   dest = type_any;        lex.Next(); break;
```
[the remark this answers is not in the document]
arvyy: I'm doing by the third comment, but what is the significance of changing `any` to `?`? I mean behavior with anytype token should be same as with a wildcard one, since it'd be looked ahead explicitly. Or do you just think <?> is better looking?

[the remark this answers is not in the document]
aardappel: I guess we currently don't have `any` as a type anymore you can specify, so I am trying to not introduce it here. Besides, it is even confusing, since `any` isn't a supertype of `int` for example. We're purely using `any` here as an escape hatch we can test for. Maybe we should be using `type_undefined` instead to be even clearer.

aardappel:
Look at the use of `tv` in your change, and its definition.

dev/src/lobster/typecheck.h
```diff
@@ -219,6 +219,23 @@ struct TypeChecker {
         }
     }
 
+    bool ConvertsToStatically(TypeRef type, TypeRef bound) {
+        if (bound == type || type->t == V_ANY)
+            return true;
+        switch (bound->t) {
```
aardappel, line 225: with 1 case, I guess you can use `if`

aardappel: You can even early out if `!= V_CLASS`

dev/src/lobster/typecheck.h
```diff
@@ -313,6 +330,20 @@ struct TypeChecker {
         return a == b;  // Not inlined for documentation purposes.
     }
 
+    bool TypeOfUndefinedGenericSpec(TypeRef t) {
+        if (!IsUDT(t->t)) return false;
```
aardappel, line 334: Just `return IsUdt && .. `

dev/src/lobster/typecheck.h
```diff
@@ -313,6 +330,20 @@ struct TypeChecker {
...
+        return IsUndefinedGenericSpec(t->udt);
+    }
+
+    bool IsUndefinedGenericSpec(UDT* udt) {
+        for (int i = 0; i < udt->generics.size(); i++) {
+            auto& utr = udt->generics[i].giventype.utr;
```
aardappel, line 340: use `&` on the right. Same with `*` above and anywhere else. Generally, keep the style of surrounding code.

dev/src/lobster/typecheck.h
```diff
@@ -219,6 +219,23 @@ struct TypeChecker {
         }
     }
 
+    bool ConvertsToStatically(TypeRef type, TypeRef bound) {
+        if (bound == type || type->t == V_ANY)
+            return true;
+        switch (bound->t) {
+            case V_CLASS:
+                if (type->t != bound->t) break;
```
aardappel, line 227: why mix `break` with `return`? maybe use `return` everywhere?

dev/src/lobster/typecheck.h
```diff
@@ -313,6 +330,20 @@ struct TypeChecker {
...
 
+    bool TypeOfUndefinedGenericSpec(TypeRef t) {
+        if (!IsUDT(t->t)) return false;
+        return IsUndefinedGenericSpec(t->udt);
+    }
+
+    bool IsUndefinedGenericSpec(UDT* udt) {
+        for (int i = 0; i < udt->generics.size(); i++) {
+            auto& utr = udt->generics[i].giventype.utr;
+            if (!utr.Null() && utr == type_undefined)
```
aardappel, line 341: when can these be null?

dev/src/lobster/parser.h
```diff
@@ -612,8 +612,13 @@ struct Parser {
                     dest = st.NewSpecUDT(dest->spec_udt->udt);
                     if (dest->spec_udt->udt->is_generic) dest->spec_udt->is_generic = true;
                     for (;;) {
-                        auto s = ParseType(false);
-                        if (st.IsGeneric(s)) dest->spec_udt->is_generic = true;
+                        UnresolvedTypeRef s;
+                        if (allow_wildcard_generic && IsNext(T_QUESTIONMARK)) {
```
aardappel, line 616: maybe lift this check out of the `for`, such that only `T<?>` is legal, but not `T<?,U>`, `T<U,?>` and other cases we can't deal with in the typechecker are illegal. That also disallows `T<?,?>` which is probably fine, the `?` can indicate multiple elements.

dev/src/lobster/typecheck.h
arvyy: hmm I used to have a this check https://github.com/aardappel/lobster/blob/4bb2e43add47955916c16a5f9b914ab08cb783d3/dev/src/lobster/typecheck.h#L711 in different place, but you're right now null isn't needed; removed

dev/src/lobster/typecheck.h
```diff
@@ -219,6 +219,22 @@ struct TypeChecker {
...
+            return true;
+        if (bound->t != V_CLASS || type->t != V_CLASS)
+            return false;
+        if (st.SuperDistance(type->udt, bound->udt) > -1)
+            return true;
+        if (!TypeOfUndefinedGenericSpec(type))
+            return false;
+        for (auto t = bound->udt; t; t = t->resolved_superclass) {
+            if (t->first == type->udt->first)
+                return true;
```
aardappel, line 233: I was about to type a long comment here explaining why checks of `SuperDistance` here might be problematic, but it got long, so instead rewrote it as a general issue here: https://github.com/aardappel/lobster/issues/99

dev/src/lobster/typecheck.h
```diff
@@ -676,7 +705,9 @@ struct TypeChecker {
                 }
                 fail:;
             }
-            TypeError("can't find specialized superclass for: " + udt.name, errn);
+            //do not fail for an undefined generic inside an `is` statement
+            if (!IsUndefinedGenericSpec(&udt))
```
aardappel, line 709: Hmm.. would be good if we don't have to check for that here somehow..

dev/src/lobster/typecheck.h
```diff
@@ -977,6 +1008,14 @@ struct TypeChecker {
                     auto t = ResolveTypeVars({ s }, errn);
                     types.push_back(&*t);
                 }
+                // in special case of using `is` with wildcard generic
+                // the type has 1 generic argument (<?>) which may
+                // mismatch required count of generic argument for that type
+                if ((types.size() == 1 && types[0]->t == V_UNDEFINED)) {
```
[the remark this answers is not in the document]
arvyy: I mean it copies that one undefined to match the expected count for the type. From what I can tell this needs to be done for a proper specialization to be created later down the block (after it's done checking the fields matching bits). Although specialization itself is rather... exceptional, I figured it'd be best to make it behave as a regular one as much as possible.

In terms of doing this in parser instead, from what I can tell I don't know at that point what the N actually should be

dev/src/lobster/typecheck.h
```diff
@@ -219,6 +219,22 @@ struct TypeChecker {
         }
     }
 
+    bool ConvertsToStatically(TypeRef type, TypeRef bound) {
+        if (bound == type || type->t == V_ANY)
+            return true;
+        if (bound->t != V_CLASS || type->t != V_CLASS)
+            return false;
+        if (st.SuperDistance(type->udt, bound->udt) > -1)
```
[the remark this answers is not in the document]
arvyy: ConvertsTo checks if it's possible to happen at runtime and checks if it's a sane downcast, while ConvertsToStatically asserts that it will be so for sure at compile time and thus looks checks if it's an upcast

dev/src/lobster/typecheck.h
```diff
@@ -219,6 +219,22 @@ struct TypeChecker {
...
+            return true;
+        if (!TypeOfUndefinedGenericSpec(type))
+            return false;
+        for (auto t = bound->udt; t; t = t->resolved_superclass) {
```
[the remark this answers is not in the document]
arvyy: this checks not by udt itself, but by `first` reference, because at this point only base type is interesting, not actual specializations. I'm not sure if/how could I make this work with SuperDistance..

dev/src/lobster/typecheck.h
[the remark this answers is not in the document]
arvyy: as per #99 I'll remove this

dev/src/lobster/typecheck.h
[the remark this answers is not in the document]
arvyy: as per #99 this will be replaced by checking first just on bound without looping over parents

dev/src/lobster/typecheck.h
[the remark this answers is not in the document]
aardappel: I thought in the parser we know how many generics the type has, so we can push that many undefineds if that makes things cleaner

aardappel:
So we can fix the semantics of `is` in this PR or another. Question is whether we still want the `T<?>` feature, given that it adds quite a bit of special purpose processing, and now that something similar can be achieved by overloading (as per commits https://github.com/aardappel/lobster/commit/96ce9f1542c0a795143eb3829e1855da4ec68ab8 and https://github.com/aardappel/lobster/commit/2a6c1ab43b220b213eaf9765276c3e3a5e8f2ee3) and discussion in https://github.com/aardappel/lobster/issues/99

PR #104, opened 2020-06-26
Fix compilation of compiled_lobster.cpp in cases when SDL is needed

stefandd:
Without the additional includes, compiled_lobster.cpp will not compile (due to SDL needing to define a SDL_main)


aardappel:
Thanks!

bandit2023:
🙅🤦💫☝️

PR #105, opened 2020-06-28
Reworked sound functions

stefandd:
Added pause, resume, and setting volume functions.

Details:

1. The existing play_xxx functions were modified to support a repeat parameter (for numbered repeats or loops); the volume parameter was removed --> set by new function sound_volume (see below).

The return value of the play_xxx functions is now 0 in case of error and the allotted channel number from 1..8 otherwise. This channel number is an input parameter for the functions sound_pause, sound_resume, and sound_volume.

2. The functions sound_pause(ch:int) and sound_resume(ch:int) pause and resume a channel, whereas the volume of a channel is set by sound_volume(ch:int, vol:float) with vol in the range 0 .. 1

3. A new sample has been provided in samples\shooter_tutorial\tut_sound.lobster (modified tut6 from the shooter tutorial) to demonstrate looped sounds, pause and resume as well as dynamically setting the volume.


dev/src/sdlaudiosfxr.cpp
```diff
@@ -417,18 +417,85 @@ void SDLSoundClose() {
...
+        return 0;
     #endif
 
     if (!SDLSoundInit())
-        return false;
+        return 0;
 
     auto snd = LoadSound(filename, sfxr);
+    int ch = -1;
     if (snd) {
```
aardappel, line 431: just do an `if (snd) return 0`, that way you don't have to start `ch` at -1, and it is all easier to read.

dev/src/sdlaudiosfxr.cpp
```diff
@@ -417,18 +417,85 @@ void SDLSoundClose() {
...
+        return false;
+
+    ch--; // SDL_Mixer enmerates channels from 0, Lobster from 1
+    Mix_Resume(ch); // fix SDL bug not clearing the paused flag on halt by resuming (even if not paused) right before halt
+    Mix_HaltChannel(ch);
+    return true;
 }
+
+bool SDLPauseSound(int ch) {
+    #ifdef __EMSCRIPTEN__
```
aardappel, line 457: lets put this `#ifdef` inside `SDLSoundInit` so its not repeated in every function.

dev/src/sdlaudiosfxr.cpp
```diff
@@ -417,18 +417,85 @@ void SDLSoundClose() {
...
+bool SDLIsPausedSound(int ch) {
+    #ifdef __EMSCRIPTEN__
+    // Distorted in firefox and no audio at all in chrome, disable for now.
+        return false;
+    #endif
+
+    if (!SDLSoundInit())
+        return false;
+    
+    return Mix_Paused(--ch) > 0;
```
aardappel, line 478: `ch - 1`

dev/src/audio.cpp
```diff
@@ -22,20 +22,72 @@ using namespace lobster;
...
+nfr("play_sfxr", "filename,loops", "SI?", "I",
     "plays a synth sound defined by a .sfs file (use http://www.drpetter.se/project_sfxr.html"
-    " to generate these). volume in range 1..128, or omit for max. returns false on error",
-    [](StackPtr &, VM &, Value &ins, Value &vol) {
-        bool ok = SDLPlaySound(ins.sval()->strv(), true, vol.True() ? vol.intval() : 128);
+    " to generate these). the default volume is the max volume (1.0)"
+    " loops is the number of repeats to play (-1 repeats endlessly, omit for no repeats)."
+    " returns the assigned channel number (1..8) or 0 on error",
+    [](StackPtr &, VM &, Value &ins, Value &loops) {
+        int ch = SDLPlaySound(ins.sval()->strv(), true, 1.0, loops.True() ? loops.intval() : 0);
```
aardappel, line 41: no need checking for `loops.True()` here? Can just use `loops.intval()`

dev/src/audio.cpp
```diff
@@ -22,20 +22,72 @@ using namespace lobster;
...
+        return Value(ch);
+    });
+
+nfr("sound_halt", "channel", "I", "B",
+    "terminates sound channel. returns false on error",
+    [](StackPtr &, VM &, Value &ch) {
+        bool ok = SDLHaltSound(ch.True() ? ch.intval() : -1);
```
aardappel, line 48: why not just pass `0` as the "no channel" value to simplify this code? then `SDLHaltSound` simply does nothing on invalid channel

dev/src/audio.cpp
```diff
@@ -22,20 +22,72 @@ using namespace lobster;
...
+        return Value(ok);
+    });
+
+nfr("sound_pause", "channel", "I", "B",
+    "pauses sound channel. returns false on error",
+    [](StackPtr &, VM &, Value &ch) {
+        int ch_idx = ch.intval();
+        bool ok = false;
+        if (ch_idx != -1)
```
aardappel, line 57: Don't think its that useful for any functions other than `play_` ones to return an error, they should simply silently do nothing on wrong channel. And init is not going to fail on these functions since they're never the first ones to be called

dev/src/audio.cpp
```diff
@@ -22,20 +22,72 @@ using namespace lobster;
...
+            ok = SDLPauseSound(ch_idx);
+        return Value(ok);
+    });
+
+/* not needed since user pauses and can track this by themselves 
+nfr("sound_ispaused", "channel", "I", "B",
```
aardappel, line 63: then delete it :)

dev/src/audio.cpp
```diff
@@ -22,20 +22,72 @@ using namespace lobster;
...
+    [](StackPtr &, VM &, Value &ch) {
+        int ch_idx = ch.intval();
+        bool ok = false;
+        if (ch_idx != -1)
+            ok = SDLResumeSound(ch_idx);
+        return Value(ok);
+    });
+
+nfr("sound_volume", "channel,volume", "IF", "B",
+    "sets the channel volume in the range 0.0..1.0. returns false on error",
```
aardappel, line 85: just say `0..1`

dev/src/sdlaudiosfxr.cpp
```diff
@@ -417,18 +417,85 @@ void SDLSoundClose() {
...
+bool SDLSetVolume(int ch, float vol) {
+    #ifdef __EMSCRIPTEN__
+    // Distorted in firefox and no audio at all in chrome, disable for now.
+        return false;
+    #endif
+
+    if (!SDLSoundInit())
+        return false;
+    
+    Mix_Volume(--ch, (int)(MIX_MAX_VOLUME*vol));
```
aardappel, line 499: spaces around `*` here and elsewhere

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
+// Extension of shooter tutorial tut7 here with background music, enemy sounds and "hot" pause
```
aardappel, line 1: put this in the same dir as the shooter tutorial.. call it `tut_sound.lobster` maybe. At the end of the shooter tutorial doc (just before all the coroutine stuff) make a paragraph saying.. want to learn how to add sound to this code? have a look at this file, and the functions starting with `play_` or `sound_`.

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
...
+    if playing:
+        // pause logic
+        if gl_button("space") == 1:
+            paused = not paused
+            if paused:
+                pausetime = t_gl
+                sound_pause(bgmusic)
+                sound_pause(ufobuzz)
+            else:
+                lastenemy  += t_gl - pausetime
```
aardappel, line 74: this time management is a bit messy. The best way to add a pause is to NOT use `gl_time` at all, but instead have a `game_time` variable which you increase with `gl_delta_time()` when the game is not paused. That then automatically freezes any animation.

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
...
+
+            let dir = xy_f { (gl_button("d") >= 1) - (gl_button("a") >= 1),
+                          (gl_button("s") >= 1) - (gl_button("w") >= 1) }
+            let newpos = playerpos + normalize(dir) * delta_time_w_pause() * playerspeed
+            if !any(abs(newpos) > float(gl_window_size()) / scale / 2):
+                playerpos = newpos
+
+            let tomouse = normalize(gl_local_mouse_pos(0) - playerpos)
+
+            if !paused:
```
aardappel, line 91: also makes this check unnecessary, since there will be no spawning if time is not advancing.

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
...
+            if paused:
+                pausetime = t_gl
+                sound_pause(bgmusic)
+                sound_pause(ufobuzz)
+            else:
+                lastenemy  += t_gl - pausetime
+                lastbullet += t_gl - pausetime
+                sound_resume(bgmusic)
+                sound_resume(ufobuzz)
+        //
```
aardappel, line 78: empty comment

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
...
+                    playerhealth -= delta_time_w_pause() * 50
+                    if playerhealth <= 0:
+                        playerhealth = 0
+                        highscore = max(highscore, score)
+                        playing = false
+                gl_color lerp(color_red, color_blue, div(e.hp, enemymaxhp)):
+                    renderpointytriangle(e.pos, playerdir)
+            if !paused:
+                // distance-dependent UFO buzz sound
+                let enemy_volume = clamp(1.0/(closest_enemy_dist*closest_enemy_dist + 0.001), 0.0, 1.0)
```
aardappel, line 126: use spaces around operators

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
...
+                gl_line_mode 1:
+                    gl_color color_grey:
+                        gl_circle(0.5, 20)
+
+            renderpointytriangle(playerpos, tomouse)
+    else:
+        let msg = "press space to play!"
+        gl_translate float(gl_window_size() - gl_text_size(msg)) / 2:
+            gl_text(msg)
+        if not paused:
```
aardappel, line 150: this is additional logic that shouldn't need to be here.

samples/soundtest.lobster
```diff
@@ -0,0 +1,173 @@
...
+import std
+import vec
+import color
+
+fatal(gl_window("Shooter Tutorial", 640, 480))
+
+check(gl_set_font_name("data/fonts/US101/US101.ttf") and gl_set_font_size(32), "can\'t load font!")
+
+// load music and sound files
+let bgmusic = play_wav("data/sounds/Cyber-Dream-Loop.wav", -1) // by Eric Matyas from https://soundimage.org/looping-music/
```
aardappel, line 12: 6MB still too big.. if there's no music that is small, lets just add an environment sound loop?

dev/src/sdlaudiosfxr.cpp
stefandd: Ok

dev/src/sdlaudiosfxr.cpp
stefandd: Ok

dev/src/sdlaudiosfxr.cpp
stefandd: Ok

dev/src/audio.cpp
stefandd: Indeed.

dev/src/audio.cpp
stefandd: Yes, should have been simply 0, but to a different effect. To SDL_mixer a channel of -1 (resulting from a 0 here) means all channels. This is what I had actually wanted: sound_halt(ch) terminates that channel; sound_halt() terminates all.

dev/src/audio.cpp
stefandd: "And init is not going to fail on these functions since they're never the first ones to be called" -- unless of course a Lobster user is calling them before calling any play. However, the behavior of the corresponding Mix_xxx functions in mixer.c is indeed benign I think so yes, it won't crash anything.

dev/src/audio.cpp
stefandd: Ok

dev/src/sdlaudiosfxr.cpp
stefandd: Will do! Wonder who wrote that code `fperiod=100.0/(p_base_freq*p_base_freq+0.001);` some lines above :D

dev/src/audio.cpp
stefandd: Ok

samples/soundtest.lobster
stefandd: Yes, it was a quick ugly hack to add hot pause to your tut7.lobster. Changed as suggested.

samples/soundtest.lobster
stefandd: Yes.

samples/soundtest.lobster
stefandd: Done

samples/soundtest.lobster
stefandd: Used now :)

samples/soundtest.lobster
stefandd: Mmmh. If you die, I want the music to stop again like it was in the beginning. This is the purpose of this code block. However, I should have init'ed paused to true.

samples/soundtest.lobster
stefandd: Ok. I hope one of us or someone will add libvorbis to the external sources and allow .ogg files (easily > 10x smaller)

docs/shooter_tutorial.html
```diff
@@ -263,6 +263,7 @@ <h3 id="whats-next">7: What’s next?</h3>
 <li><p>Better UI and main menu.</p></li>
 <li><p>Game structure: different levels the player can try to beat, permanent upgrades the player can buy, etc.</p></li>
 </ul>
+<p>Also, at this point, our game has no sound at all. If you'd like to see how to add some anbient or enemy sounds, have a look at the use of the <code>play_</code> and <code>sound_</code> functions in <code>tut_sound.lobster</code>. This example also shows how to to dynamically adjust the volume and pause or resume individual sounds.</p>
```
aardappel, line 266: you'll want to change this in the .md, it will get overwritten otherwise

stefandd: Ah -- wasn't even aware of the source files. Done.

aardappel:
Awesome, thanks!

PR #106, opened 2020-06-29
Further tweaks to sound-related update

stefandd:
- small fix in audio.cpp to use .fltval() instead of .fval()
- added distance-dependent projectile impact sound to samples/shooter_tutorial/tut_sound.lobster
- added AUTHORS.TXT to data/sounds to satisfy attribution license requirements


aardappel:
Nice!

PR #107, opened 2020-06-29
Added unix epoch to the vector returned by date_time()

stefandd:
This will allow unique random numbers for each program start simply by using `rnd_seed(date_time()[0])`


aardappel:
Can you update `date_time.lobster` ?

aardappel:
Thanks!

PR #109, opened 2020-07-07
Fix the [any] versions of the any() and "all() builtins

stefandd:
In cases when different versions of a function for numerical structs and an [any] inputs exist, it seems that [int] vectors are not processed properly. For the builtins `any` and `all` this is fixed by simply switching the definitions so that the [any] versions are registered and then processed first.


aardappel:
Thanks! Really confusing diff..

PR #111, opened 2020-07-10
Extend the builtin `string_to_int` to support different bases, enabling conversion of hexadecimal const strings

stefandd:
Simple extension of `string_to_int(s:string, base:int) -> int, int` to support bases other than 10. This complements the already existing support for this in `number_to_string` and is very useful to convert string constants read from an external source (html color tables etc.) The base parameter is optional and defaults to 10, so no existing code will break.

Old:
`print string_to_int("0xffff")` prints `0`.

Now:
`print string_to_int("0xffff", 16)` prints `65535`.

### DIFF
--- a/dev/src/builtins.cpp
+++ b/dev/src/builtins.cpp
@@ -354,7 +354,7 @@ nfr("substring", "s,start,size", "SII", "S",
         return Value(ns);
     });
 
-nfr("string_to_int", "s", "SI?", "IB",
+nfr("string_to_int", "s,base", "SI?", "IB",
     "converts a string to an int given the base (2..36, e.g. 16 for hex, default is 10)."
     "returns 0 if no numeric data could be parsed; second return value is true if all"
     "characters of the string were parsed.",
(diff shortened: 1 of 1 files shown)


dev/src/builtins.cpp
```diff
@@ -354,13 +354,17 @@ nfr("substring", "s,start,size", "SII", "S",
         return Value(ns);
     });
 
-nfr("string_to_int", "s", "S", "IB",
-    "converts a string to an int. returns 0 if no numeric data could be parsed."
-    "second return value is true if all characters of the string were parsed",
-    [](StackPtr &sp, VM &, Value &s) {
+nfr("string_to_int", "s", "SI?", "IB",
```
aardappel, line 357: note the `"s"` which are the names of the arguments, used in the automatic help generation. You added an argument, so this should probably be `"s,base"`. If you look at the comment above the assert that failed in CI, it says so :)

stefandd: Oops, I guess I was too quick. Will fix

aardappel:
Thanks!

PR #112, opened 2020-07-15
Added new sound_channels builtin

stefandd:
This adds a new sound command `sound_channels() -> int` by which the number of currently still available channels is obtained. This is useful in situations when more sounds are potentially generated than channels are available, i.e. when for example many enemy ships shoot lasers, potentially filling up the entire mixer channel range. With this command, a Lobster program can check how many channels are still free and decide to not add another sound of a certain type in order to reserve a channel for a more prioritized sound effect (player ship etc.)

### DIFF
--- a/dev/src/audio.cpp
+++ b/dev/src/audio.cpp
@@ -27,7 +27,7 @@ nfr("play_wav", "filename,loops,prio", "SI?I?", "I",
     " will automatically be converted on first load). the default volume is the max volume (1.0)"
     " loops is the number of repeats to play (-1 repeats endlessly, omit for no repeats)."
     " prio is the priority of the sound which determines whether it can be deleted or not"
-    " in case of too many play function calls."
+    " in case of too many play function calls (defaults to 0)"
     " returns the assigned channel number (1..8) or 0 on error",
     [](StackPtr &, VM &, Value &ins, Value &loops, Value &prio) {
         int ch = SDLPlaySound(ins.sval()->strv(), false, 1.0, loops.intval(), prio.intval());
@@ -39,17 +39,10 @@ nfr("play_sfxr", "filename,loops,prio", "SI?I?", "I",
     " to generate these). the default volume is the max volume (1.0)"
     " loops is the number of repeats to play (-1 repeats endlessly, omit for no repeats)."
     " prio is the priority of the sound which determines whether it can be deleted or not"
-    " in case of too many play function calls."
+    " in case of too many play function calls (defaults to 0)"
     " returns the assigned channel number (1..8) or 0 on error",
     [](StackPtr &, VM &, Value &ins, Value &loops, Value &prio) {
         int ch = SDLPlaySound(ins.sval()->strv(), true, 1.0, loops.intval(), prio.intval());
-        return Value(ch);
-    });
-
-nfr("sound_channels", "", "", "I",
-    "returns the number of currently available sound channels.",
-    [](StackPtr &, VM &) {
-        int ch = SDLAvailChannels();
         return Value(ch);
     });
 
--- a/dev/src/audio.cpp
+++ b/dev/src/audio.cpp
@@ -42,6 +42,13 @@ nfr("play_sfxr", "filename,loops", "SI?", "I",
         return Value(ch);
     });
 
+nfr("sound_channels", "", "", "I",
+    "returns the number of currently available sound channels.",
+    [](StackPtr &, VM &) {
+        int ch = SDLAvailChannels();
+        return Value(ch);
+    });
+
 nfr("sound_halt", "channel", "I", "",
     "terminates a specific sound channel.",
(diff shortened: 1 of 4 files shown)


aardappel:
Doesn't SDL_Mixer handle this automatically already? By default, it has 8 channels, so any sounds attempted to be played while there's already 8 playing will get dropped, which sounds like the right behavior for a simple game.

If anything, we should be calling `Mix_AllocateChannels(16)` after `Mix_OpenAudio` at least, since 8 seems a bit low to me.. that should reduce the chances of sounds dropping too.

If we wanted to make a fancier sound interface, then `play_sound` should take a priority parameter, and then if a sound fails to play because channels are full, it finds the lowest pri sound, kills it if its lower than the current sound, and tries again. That way a "game over" sound would not be dropped because you have 8 bullet sounds playing. In fact I'm surprised SDL_Mixer doesn't already have this functionality.

stefandd:
Extended pre-allocated channels to 16; added prioritization of sounds as an argument to the play_xxx functions, and added commands to interrogate the status of a channel (playing, paused, available) etc. The prioritization scheme is that 3 priority levels are supported: low/normal (0), elevated (1), and high (2). Adding a sound will check for a free channel and if none is available will delete the oldest sound effect from the lowest possible priority (including up to the same priority as the caller)

dev/src/sdlaudiosfxr.cpp
```diff
@@ -422,14 +424,30 @@ void SDLSoundClose() {
...
+        for (int p = 0; p <= prio; p++) {
+            ch_try = Mix_GroupOldest(p);
+            if (ch_try > -1) {
+               Mix_HaltChannel(ch_try); 
+               ch = ch_try;
+               break;
+            }
+        }
+    }
+    if (ch > -1) { // ch should never be -1 at this point anyhow
```
aardappel, line 446: `ch >= 0`

ch might be -1 here if you tried to play something with normal prio, and all channels are taken up by higher prio groups.

dev/src/audio.cpp
```diff
@@ -22,31 +22,54 @@ using namespace lobster;
...
+    " in case of too many play function calls."
     " returns the assigned channel number (1..8) or 0 on error",
-    [](StackPtr &, VM &, Value &ins, Value &loops) {
-        int ch = SDLPlaySound(ins.sval()->strv(), true, 1.0, loops.intval());
+    [](StackPtr &, VM &, Value &ins, Value &loops, Value &prio) {
+        int ch = SDLPlaySound(ins.sval()->strv(), true, 1.0, loops.intval(), prio.intval());
         return Value(ch);
     });
 
+nfr("sound_channels", "", "", "I",
```
aardappel, line 49: I don't think we want this and the following function.. I'd rather sound management be full automatic once you choose good priorities

dev/src/sdlaudiosfxr.cpp
```diff
@@ -422,14 +424,30 @@ void SDLSoundClose() {
...
-
     auto snd = LoadSound(filename, sfxr);
     if (!snd) return 0;
-
-    int ch = Mix_PlayChannel(-1, snd->chunk.get(), loops);
-    if (ch >= 0) Mix_Volume(ch, (int)(MIX_MAX_VOLUME * vol)); // set channel to default volume (max)
+    if (prio < 0) prio = 0;
+    else if (prio > 2) prio = 2;
+    int ch = Mix_GroupAvailable(-1); // is there any free channel?
+    if (ch == -1) { // no free channel?
```
aardappel, line 434: `ch < 0`

dev/src/sdlaudiosfxr.cpp
```diff
@@ -422,14 +424,30 @@ void SDLSoundClose() {
...
+            if (ch_try > -1) {
+               Mix_HaltChannel(ch_try); 
+               ch = ch_try;
+               break;
+            }
+        }
+    }
+    if (ch > -1) { // ch should never be -1 at this point anyhow
+        Mix_PlayChannel(ch, snd->chunk.get(), loops);
+        Mix_GroupChannel(ch, prio); // add the channel to the group with tag prio
```
aardappel, line 448: Here we assign a channel to a group.. how does it ever get de-assigned? in particular, any sounds in the highest prio group that finishes playing should immediately be available to the any other prio group.. how does that currently work?

If this gets complicated, I think managing prios without groups would be better

aardappel:
Ok, looks good now, thanks!

dev/src/audio.cpp
stefandd: I do agree re: `sound_channels`. Regarding the `sound_status` function, isn't that useful to be able to get the channel (is it paused/playing etc.)?

aardappel: I'm not sure when that would be useful, but left it in for now

PR #114, opened 2020-07-16
Tweak to the channel selection algorithm to include sound sample age

stefandd:


aardappel:
Cool, exactly what I had in mind :)

PR #123, opened 2020-12-11
Add a check for SELinux

autumn-birds:
Previously, lobster was silently failing to run code on my system, which happened to come with SELinux enabled. This should fix the problem (it seems to work on my system, I have not personally tested on anything else.)

I imagine it might be a good idea to check for and report this error somehow in lobster itself, but I didn't manage to do that in this change. Maybe eventually, who knows.

It also won't fix the prebuilt binaries, I'm not really sure what to do about that.

aardappel patiently helped out with suggestions and spotting a silly typo.


autumn-birds:
Oh, it looks from these checks like it doesn't actually build on... systems without `selinuxenabled`. Oof. I don't know how to fix my fudged-together change to do that.

aardappel:
Maybe remove the `REQUIRED` from the `find_program` ?

It's weird though, on my ubuntu I don't have `selinuxenabled` installed, but `find_program` doesn't complain (just makes `SELINUX_CHECKER` false).

autumn-birds:
Wow, I must have been really tired or something when I was doing this -- I was worried this exact thing would happen and somehow I put `REQUIRED` anyway.

aardappel:
Awesome, thanks!

PR #124, opened 2020-12-23
Replace #defines for Win compilation

stefandd:
First step to re-enable MingW/GCC support on Windows


aardappel:
perfect thanks!

PR #125, opened 2020-12-24
Even more MingW fixes

stefandd:


aardappel:
Thanks!
Weird, these didn't show up when I built it.. I guess slightly different gcc version?

PR #126, opened 2021-03-06
Add some missing SDL files to Windows .sln

sgraham:
I had to add these to compile using the checked in Windows .sln, or otherwise got link errors (VS2019).


aardappel:
I haven't had that problem yet, I wonder what the difference is. But let's merge for now.
Thanks!

PR #128, opened 2021-04-25
Fix crash building error message for function arity mismatch.

MortimerSnerd:
On linux, you can get a segmentation fault when you have a program
that passes the wrong number of parameters to a function.

The crash was occuring because Lex::Error was being passed an
invalid "ln" parameter from the Parser::ErrorAt() function.
ErrorAt() was taking the address of "what->line" when "what"
was null.

Fixes WarnAt() in passing, as it had the same pattern.


aardappel:
Thanks! I recently refactored that code.. apparently badly!

aardappel:
That reminds me we should really have a way to test error messages..

PR #129, opened 2021-05-14
Fix for eating bread not reducing hunger.

MortimerSnerd:
The goaptest.lobster works fine, but I noticed while trying to change it to force other outcomes that it would not buy flour and make bread when I changed the states and costs. 

Ends up being a syntax problem, where the action of "eat" was accidentally including the "s[hungry]--" in the else block.


aardappel:
Wow, what a find! Thanks :)

PR #132, opened 2021-05-20
Assertion failure when importing module with a private enum.

MortimerSnerd:
The assertion in SymbolTable::Unregister(Enum*, ...) was failing on a private enum, because the enum and its values had already been removed by EndOfInclude -> ErasePrivate(enums).    I just changed the ordering of the checks so it only tries to remove the enum values if the enum is found in the passed `dict`.


MortimerSnerd:
Hold up, I must have made a mistake when I rebased the branch.  I don't think I can fix in GitHub UI, I'll close and re-do with a proper branch.

PR #133, opened 2021-05-20
Fix asertion when module with private enum is imported.

MortimerSnerd:
When a private enum was being Unregistered by Parser::CleanupStatements,
the assertion in SymbolTable::Unregister(Enum*, ...) was failing
because the enum's values could not be found.  This is because the
private enum had already been unregistered by the call chain
Parser::ParseStatements() -> SymbolTable::EndOfInclude() ->
ErasePrivate(enums).

The fix is to re-arrange the conditionals so the values aren't
erased if the enum has already been erased.


aardappel:
Makes sense, thanks for the fix!

PR #135, opened 2021-06-05
Adds clipping rectangles state

MortimerSnerd:
Not a bug, but I found a place with some text handling where it would be convenient
to clip to a rectangle, so I added some code for it in a branch.

- Adds gl_scissor_push() and gl_scissor_pop() primatives to push and pop
  clipping rectangles.  The clipping rectangle state is cleared
  at the beginning of each frame, so a missing pop call won't foul
  up subsequent frames.
- Adds a gl_scissor(x, y, w, h, body) function that enables a clipping
  rectangle while body() is active.  (barring return from shennanigans)

Feel free to close this out if it's not something you want or need in the base image.

### DIFF
--- a/dev/src/glsystem.cpp
+++ b/dev/src/glsystem.cpp
@@ -33,6 +33,8 @@ float custompointscale = 1.0f;
 bool mode2d = true;
 bool mode_srgb = false;
 GeometryCache *geomcache = nullptr;
+vector<int4> scissorRects;
+int2 viewportSize;
 
 BlendMode SetBlendMode(BlendMode mode) {
     static BlendMode curblendmode = BLEND_NONE;
@@ -70,6 +72,30 @@ BlendMode SetBlendMode(BlendMode mode) {
 void ClearFrameBuffer(const float3 &c) {
     GL_CALL(glClearColor(c.x, c.y, c.z, 1.0));
     GL_CALL(glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT));
+}
+
+void SetScissorRect(int4 scis)
+{
+   // Always get bitten by this, glScissor x & y are BOTTOM left, not top left.
+   GL_CALL(glScissor((GLint)scis.x, (GLint)(viewportSize.y - (scis.y + scis.w)), 
+                     (GLint)scis.z, (GLint)scis.w));
+   GL_CALL(glEnable(GL_SCISSOR_TEST));
+}
+
+void PushScissorRect(int4 scis)
+{
+   SetScissorRect(scis);
+   scissorRects.push_back(scis);
+}
+
+void PopScissorRect()
+{
+   scissorRects.pop_back();
+   if (!scissorRects.empty()) {
+      SetScissorRect(scissorRects.back());
+   } else {
+      GL_CALL(glDisable(GL_SCISSOR_TEST));
+   }
 }
 
 void Set2DMode(const int2 &ssize, bool lh, bool depthtest) {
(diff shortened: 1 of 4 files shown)


dev/src/graphics.cpp
```diff
@@ -179,6 +179,24 @@ nfr("gl_load_materials", "materialdefs,inline", "SI?", "S?",
         return err[0] ? Value(vm.NewString(err)) : NilVal();
     });
 
+nfr("gl_push_scissor", "x,y,width,height", "IIII", "",
```
aardappel, line 182: use actual vector types, so "I}:2I}:2" is two 2D int vectors.

dev/src/glsystem.cpp
```diff
@@ -33,6 +33,8 @@ float custompointscale = 1.0f;
 bool mode2d = true;
 bool mode_srgb = false;
 GeometryCache *geomcache = nullptr;
+vector<int4> scissorRects;
+int2 viewportSize;
```
aardappel, line 37: No need to duplicate that state here, just call `GetScreenSize()`

dev/src/graphics.cpp
```diff
@@ -179,6 +179,24 @@ nfr("gl_load_materials", "materialdefs,inline", "SI?", "S?",
...
+    "Enables scissor testing so only the pixels in the given rectangle can "
+    "be written.  (x,y) is the top left corner. Scissor state only lasts till "
+    "the end of the frame.",
+    [](StackPtr &, VM &vm, Value &x, Value &y, Value &w, Value &h) {
+   TestGL(vm);
+   PushScissorRect(int4(x.ival(), y.ival(), w.ival(), h.ival()));
+   return NilVal();
+});
+
+nfr("gl_pop_scissor", "", "", "",
```
aardappel, line 192: while this push & pop is the most "correct" design, it is not the simplest (needing a stack of values), and for such niche functionality, I prefer the simpler design of some of the other GL functions. I would suggest a single function that a) turns scissor OFF is size == screensize, otherwise on, and b) returns the previous existing scissor size if enabled (`glGet(GL_SCISSOR_BOX..)`), or returns the full screen size if disabled. That way, the code in `gl.lobster` can be the same as for `gl_color` etc.

MortimerSnerd: Fair enough.  I didn't end up nesting them anyway.  I've re-implemented in those terms, and cleaned up the style and indent problems.

MortimerSnerd: Also, I didn't change this because I didn't have a good idea for a test case, but it looks like the PopVec() calls for gl_start_text_input() are out of order.

aardappel: well, my design still allows nesting, just the "stack" for nesting will be taken care of by Lobster code instead of C++ code :)

dev/src/glsystem.cpp
```diff
@@ -72,6 +73,34 @@ void ClearFrameBuffer(const float3 &c) {
...
+        prev_size.x = curbox[2];
+        prev_size.y = curbox[3];
+    } else {
+        prev_topleft.x = 0;
+        prev_topleft.y = 0;
+        prev_size = scrnsz;
+    }
+
+    // Always get bitten by this, glScissor x & y are BOTTOM left, not top left.
+    GL_CALL(glScissor((GLint)topleft.x, (GLint)(scrnsz.y - (topleft.y + size.y)),
```
aardappel, line 95: only call this if you're enabling it

dev/src/glsystem.cpp
```diff
@@ -72,6 +73,34 @@ void ClearFrameBuffer(const float3 &c) {
     GL_CALL(glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT));
 }
 
+void SetScissorRect(int2 topleft, int2 size, int2& prev_topleft, int2& prev_size) {
+    int2 scrnsz = GetScreenSize();
+    GLboolean enabled;
+
+    GL_CALL(glGetBooleanv(GL_SCISSOR_TEST, &enabled));
+    if (enabled) {
+        GLint curbox[4];
```
aardappel, line 82: make this function take a `pair<int2, int2> &prev` instead, then you can call `glGetBooleanv(GL_SCISSOR_TEST, prev.first.data())` and avoid all the copying. And below you can then `prev = { int2_0, scrnsz }`

aardappel:
Nice!

PR #141, opened 2021-08-06
Modify cmake config to static link gcc libraries

franko:
When in release mode and using gcc instruct cmake to static link libgcc and libstdc++.

If used in with tdm-gcc it allows the creation of a self-contained lobster executable
that does not require additional shared libraries.


aardappel:
I'm a fan of static linking, so sounds good to me.. thing is though, this matches any GCC setup, would this work similarly on Linux?

franko:
It works perfectly on linux too and it does exactly the same thing, it links using the static version of the libgcc and libstdc++. I am pretty much sure it also works on macOS but I haven't tested because I use clang on macOS.

I always use this config in all my projects because it is annoying to me to have to provide the share libraries separately.

aardappel:
Ok!

PR #144, opened 2021-08-09
Von koch fractal sample application

franko:
Nothing really special but here it is, my favorite fractal curve implemented in Lobster.

### DIFF
--- a/samples/vonkoch.lobster
+++ b/samples/vonkoch.lobster
@@ -1,6 +1,7 @@
-import vec
 import color
 import gl
+import imgui
+import vec
 
 // Sequence of the angles for the Von Koch curve's segment. A line
 // will be drawn along each angle by continuing from the previous point.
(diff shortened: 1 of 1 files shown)


aardappel:
I guess the way imgui is used is nice because it is super simple, but there is little use to have depth below 5 (looks undetailed) or above 6 (runs slow).

`triangle_seq` could be inline, or simply `i * 120` in the loop.

`curr_angle` could be `seq[max(0, i - 1)]`, after which the `gl_rotate_z` call could be moved to inside `curve` such that the duplication between the two `for` loops can be removed.. or you could make the recursion start with the `for` loop, and make the `if` `else` be below that to remove the duplication.

aardappel:
Ah sorry, had my Lobster in debug mode, in release it only gets slow at depth 8 :) But still, even in 4K, can't see any added detail in 7 above what I see in 6..

franko:
I will try to do the simplifications you suggested, thank you.

I thought that adding imgui was nice to show people how simple it is to add a control and also to show how the fractal "works".

PR #147, opened 2021-08-11
Get CMake working on macOS

dcurrie:
Move include for `external/libtcc` to `CMAKE_CXX_FLAGS` to exclude it from SYSTEM includes by using `-iquote`.

### DIFF
--- a/dev/CMakeLists.txt
+++ b/dev/CMakeLists.txt
@@ -147,7 +147,10 @@ include_directories(include
                     external/SDLMixer
                     external/freetype/include
                     external/imgui
-                    external/libtcc)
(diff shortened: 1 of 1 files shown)


aardappel:
Hmpf just realized that this will not allow the CMake file to work on Visual Studio anymore, and though that is not used at the moment, would be good to have that keep working.

Do you mind adding an `if (CMAKE_GENERATOR MATCHES "Visual Studio")` around the `set` and then adding the dir normally?

aardappel:
great, thanks!

aardappel:
Reverted: https://github.com/aardappel/lobster/commit/5b94e130067890ccd171902efceb176b013156b9

PR #154, opened 2021-11-13
Expand box2d API

acgaudette:
this PR implements a few body functions from box2d under the ph namespace:
```
ApplyLinearImpulseToCenter => ph_apply_linear_impulse_to_center (emulated)
SetLinearVelocity          => ph_set_linear_velocity
GetMass                    => ph_get_mass
```
and should serve as an example for future changes of a similar nature?

### DIFF
--- a/dev/src/physics.cpp
+++ b/dev/src/physics.cpp
@@ -205,6 +205,17 @@ nfr("ph_dynamic", "shape,on", "RB", "",
         return NilVal();
     });
 
+nfr("ph_set_linear_velocity", "id,velocity", "RF}:2", "",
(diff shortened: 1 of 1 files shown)


aardappel:
Thanks, yes these bindings were very incomplete so this is welcome :)

aardappel:
Generally, the Box2D API is very big, so I've only added bindings for things that I know I have a need for, since bindings for every function would probably be too low level. The other thing is to see if things can be simplified, i.e. if things that take multiple API calls in C/C++ can become a single simpler Lobster built-in.

PR #155, opened 2021-11-13
Convert vscode syntax to .json

dbaileychess:
Converts the tmLanguage XML to json for a bit less verbose format. I used the extension in https://stackoverflow.com/questions/61283282/how-to-convert-a-textmate-grammar-xml-flavor-to-either-yaml-or-json-flavor to do the conversion.

Added `*.code-workspace` to gitignore so that I can define a [multi-root](https://code.visualstudio.com/docs/editor/multi-root-workspaces) project for Lobster.

Added `docs/vscode/lobster/.vscode/launch.json` which (when combined with multi-root project) allows launching the extension debugger in vscode using 'F5'.


aardappel:
Awesome, thanks!

PR #156, opened 2021-11-14
Updated vscode syntax to cover most of the language

dbaileychess:
Updated the syntax for vscode to include most of the language features.

There is still a bunch to do (`generics`, `fn`, `specializers` etc) but this PR covers most things well.

Added a new example file that tries to cover all the grammar cases, so that it is easy to see if the syntax highlighting is working over the whole language.


samples/syntax_highlight.lobster
```diff
@@ -0,0 +1,134 @@
+// This is an example Lobster file to demonstrate syntax highlighting.
```
aardappel, line 1: can we move this file to `tests` ?

aardappel (commented):
Nice.. thanks for being so thorough :)

samples/syntax_highlight.lobster
dbaileychess: Done!

dbaileychess:
I thought it was a good way to learn the language, at least the grammar.

PR #159, opened 2022-02-10
Long lexer strings now point to their token type string

Pombal:
This is so that they only have to be modified in one place ( ttypes.h ).

I understand this is less readable and makes it uneven with the 1 character tokens so feel free to not merge.


aardappel:
Yes, that's a nice cleanup, it looks like you'd need to change `"include"` in `ttypes.h` to `"import"` for this to work. So yes, the names had already diverged :)

Pombal:
Yep, I just did that, good catch. I need to get myself setup to run the unit tests locally.

aardappel:
Thanks!

Issue #160, opened 2022-02-18
Shooter tutorial fails at Step 3


sgraham:
I was just following http://aardappel.github.io/lobster/shooter_tutorial.html to get a feel for Lobster and noticed that on "Step 3" it doesn't match the desired output (the "ship" disappears).

Taking this as an opportunity to see how debugging would go :) I believe it's because when no WASD is pressed `dir` is a zero-vec which `normalize` then turns into NaNs.

It looks like this might have broken fairly recently in https://github.com/aardappel/lobster/commit/b9e79da6579dcf9d91d076837f35c503bdf17105#diff-7cc745b151e18e0112b4f542bc3566974bacb4448af236d24c7a50ecab856a83R706 as previously a zero-length vector was maintained by `normalize`.

Not sure if it makes sense for `normalize` to special case 0... I guess the tutorial code should probably check instead. (A debugging mode that panics on NaNs might be a nice wishlist thing too. :)

aardappel:
Wow, nice you found the root cause!

In the past, Lobster was a more high level language than today, and this special casing made sense.
It is a good question what really it should be doing ideally.
If you do an actual scalar divide by zero, you get a runtime error... certainly I don't think a NaN here is desirable.
So it may make sense that normalize should follow this pattern and runtime error, forcing the programmer to check.

That said, for vector ops, Lobster tries to follow glsl where possible, and there implementations just return a zero vector. That is kind of useful, as you see here.
The reason the scalar divide doesn't just give a zero result is because it may mask bugs, and it is not clear that zero would be a good default anyway (infinity makes more sense), but in the case of normalize, which is used typically for graphics, a zero vector (or color) actually is likely much more helpful.

So I'd vote to keep the previous behavior :)
Do you want to make a PR, since you found it?

PR #161, opened 2022-02-18
Restore normalize allowing passthrough of zero vec

sgraham:
Possible fix for #160 (not sure if there's a more efficient way of doing this).

### DIFF
--- a/dev/src/builtins.cpp
+++ b/dev/src/builtins.cpp
@@ -710,7 +710,11 @@ nfr("normalize", "vec",  "F}" , "F}",
         VECTORVARS;
         VECTOROPNR(sql += f.fval() * f.fval());
         double m = sqrt(sql);
-        VECTOROPNR(elems[i] = f.fval() / m);
+        if (m == 0.0) {
+            VECTOROPNR((void)f; elems[i] = 0.0);
+        } else {
+            VECTOROPNR(elems[i] = f.fval() / m);
+        }
     });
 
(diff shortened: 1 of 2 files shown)


sgraham:
Oops, some builds don't like that. I should find a place to put a test too.

dev/src/builtins.cpp
```diff
@@ -710,7 +710,13 @@ nfr("normalize", "vec",  "F}" , "F}",
         VECTORVARS;
         VECTOROPNR(sql += f.fval() * f.fval());
         double m = sqrt(sql);
-        VECTOROPNR(elems[i] = f.fval() / m);
+        if (m == 0.0) {
+            for (iint i = 0; i < len; i++) {
```
aardappel, line 714: `VECTOROPNR(elems[i] = 0.0)` ?

sgraham: That's what I did at first, but the gcc/clang complained that ```/home/runner/work/lobster/lobster/dev/src/builtins.cpp:584:14: error: variable ‘f’ set but not used [-Werror=unused-but-set-variable]
  584 |         auto f = elems[i]; \```

Maybe `VECTOROPNR((void)f; elems[i] = 0.0)` if you prefer that?

aardappel: yeah, that's not great, but still better to be consistent.

sgraham: Done

PR #166, opened 2022-03-23
Empty dictionary with out of bounds assertion on get() call.

MortimerSnerd:
"buckets" is an empty array until the first set call allocates it.
So a get() before any sets will fail.

Changed set() to check the buckets length, and just return the default
value if the bucket index > length(buckets).


aardappel:
Thanks!
This is a good solution for the moment, but really I need to make construction more flexible such that this vector can be constructed earlier, so we don't need to have two unnecessary if-thens in both these methods.

PR #174, opened 2022-05-20
Use static_assert where possible in vec<>

bartwe:
Introduce vec_iterator for easy iteration of vec<> ranges
Introduce in_range and within_bounds

### DIFF
--- a/dev/src/lobster/geom.h
+++ b/dev/src/lobster/geom.h
@@ -413,11 +413,6 @@ template<typename T, int N> inline bool in_range(const vec<T, N> &v, const vec<T
     DOVECB(true, _ && v[i] >= bias.c[i] && v[i] < (bias.c[i] + range.c[i]));
 }
 
-template<typename T, int N> inline bool within_bounds(const vec<T, N> &v, const vec<T, N> &lo,
-                                                   const vec<T, N> &hi) {
-    DOVECB(true, _ && v[i] >= lo.c[i] && v[i] < hi.c[i]);
-}
-
 template<typename T, int N, typename R> inline vec<float, N> rndunitvec(RandomNumberGenerator<R> &r) {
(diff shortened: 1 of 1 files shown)


dev/src/lobster/geom.h
```diff
@@ -349,6 +408,16 @@ template<typename T, int N> inline vec<T, N> clamp(const vec<T, N> &v, T lo, T h
     DOVECR(clamp(v.c[i], lo, hi));
 }
 
+template<typename T, int N> inline bool in_range(const vec<T, N> &v, const vec<T, N> &range,
+                                                   const vec<T, N> &bias = vec<T, N>(0)) {
+    DOVECB(true, _ && v[i] >= bias.c[i] && v[i] < (bias.c[i] + range.c[i]));
+}
+
+template<typename T, int N> inline bool within_bounds(const vec<T, N> &v, const vec<T, N> &lo,
```
aardappel, line 416: these do the same, lets only have `in_range` ?

dev/src/lobster/geom.h
```diff
@@ -242,6 +242,65 @@ template<typename T, int N> struct vec : basevec<T, N> {
...
+
+    vec<T, N> &operator*() {
+        return value;
+    }
+    vec<T, N> *operator->() {
+        return &value;
+    }
+
+    vec_iterator &operator++() {
+        for (int i = 0; i < N; ++i) {
```
aardappel, line 262: This iterator is certainly neat, and avoids the usual 3 nested for loops I have a bunch of thruout the code. But I'd be warey to use it in any case where the actual speed of the iteration is likely a factor in the overall speed (very small body of the innerloop), since even though this `operator++` typically does not go thru the `for` more than once, its still more complicated logic the compiler can't completely get out of the way of the innermost loop. Probably fine in most cases though.

PR #177, opened 2022-05-29
wether -> whether

AntonBogun:
wether
_/ˈwɛðə/_

Noun
> a castrated ram.

### DIFF
--- a/dev/external/flatbuffers/src/idl_gen_text.cpp
+++ b/dev/external/flatbuffers/src/idl_gen_text.cpp
@@ -266,7 +266,7 @@ struct JsonPrinter {
           fd.value.offset);
     } else if (fd.flexbuffer && opts.json_nested_flexbuffers) {
       // We could verify this FlexBuffer before access, but since this sits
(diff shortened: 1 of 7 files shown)


dev/external/flatbuffers/src/idl_gen_text.cpp
```diff
@@ -266,7 +266,7 @@ struct JsonPrinter {
           fd.value.offset);
     } else if (fd.flexbuffer && opts.json_nested_flexbuffers) {
       // We could verify this FlexBuffer before access, but since this sits
-      // inside a FlatBuffer that we don't know wether it has been verified or
+      // inside a FlatBuffer that we don't know whether it has been verified or
```
aardappel, line 269: this file is copied from the upstream FlatBuffers repo, so unless it is fixed there, will be undone again.

aardappel:
Thanks!

dev/external/flatbuffers/src/idl_gen_text.cpp
AntonBogun: the wether shall live on

PR #179, opened 2022-06-05
Improve hashing of vector

bartwe:
by avoiding having the same hashcode for sequences with repeats that cancelled by making order significant.


aardappel:
Thanks!
FYI: most of the codebase still uses C-style casts.

PR #181, opened 2022-06-18
Docs: Updated requirements to VS2022

DerTee:
building with VS2019 failed because build tools v143 are required which
are only available for VS2022


aardappel:
Thanks for the fixes!

PR #183, opened 2022-06-20
Allow gl_perspective to override viewport 

bartwe:


dev/src/glvr.cpp
```diff
@@ -181,7 +181,7 @@ void VREye(int eye, float znear, float zfar) {
     SwitchToFrameBuffer(mstex[eye], GetScreenSize(), true, mstf, retex[eye]);
     auto proj =
         FromOpenVR(vrsys->GetProjectionMatrix((vr::EVREye)eye, znear, zfar));
-    Set3DMode(80, 1, znear, zfar);
+    Set3DMode(80, int2_1, znear, zfar);
```
aardappel, line 184: this seems to be the hacky part.. instead of an aspect ratio of `1` were now passing pixel size of `(1,1)` as magical value? Why not pass `rtsize` and remove the special case if-check above for the same result?

aardappel:
Also this is the Lobster repo which has nothing to do with ray-tracing in specific :)

dev/src/graphics.cpp
```diff
@@ -641,18 +641,21 @@ nfr("gl_line", "start,end,thickness", "F}F}1F", "",
...
     " 60), far plane (furthest you want to be able to render, try 1000) and near plane (try"
     " 1). Optionally specify a framebuffer size to override the current gl_framebuffer_size",
     [](StackPtr &sp, VM &) {
+        int2 fbo = Top(sp).True()
+            ? PopVec<int2>(sp)
+            : (Pop(sp), int2_0);
         int2 fbs = Top(sp).True()
             ? PopVec<int2>(sp)
-            : (Pop(sp), GetFrameBufferSize(GetScreenSize()));
+            : (Pop(sp), GetFrameBufferSize(GetScreenSize()) - fbo);
```
aardappel, line 654: `- fbo` is redundant: if the fbs is not specified, the fbo isn't either, and thus is 0. I guess can leave it in if we ever will have the ability to do positional argument builtin calls? :)

PR #184, opened 2022-07-06
Don't reset gl_button frame counter on repeats

Dheu3:
Most game code doesn't care about repeats. But they do care about when a button is pressed versus when it is held, and for how long. Repeats reset the counter, making it seem like a button was pressed when it wasn't. If a program actually wants repeats it can use gl_key_repeat or use the gl_button counter.

### DIFF
--- a/dev/src/sdlsystem.cpp
+++ b/dev/src/sdlsystem.cpp
@@ -62,7 +62,7 @@ struct KeyState {
 
     void Set(bool on) {
         if (on) {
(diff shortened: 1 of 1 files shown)


aardappel:
Did you test that this change actually does anything? Because `Set` (where your change is) only gets called from `updatebutton` if `repeat` is false.

Dheu3:
Thanks, looks like it was fixed recently already. I'll update

PR #188, opened 2022-07-19
Async stdout read in launch_subprocess

AntonBogun:
Fixes deadlock by filling up stdout pipe buffer.
However, will still deadlock when filling stderr pipe buffer - a proper solution would require an update in subprocess.h.

### DIFF
--- a/dev/src/platform.cpp
+++ b/dev/src/platform.cpp
@@ -568,7 +568,7 @@ void SetConsole(bool on) {
     #endif
 }
 
-iint LaunchSubProcess(const char** cmdl, const char* stdins, string& out) {
+iint LaunchSubProcess(const char **cmdl, const char *stdins, string &out) {
     #ifndef PLATFORM_ES3
         struct subprocess_s subprocess;
         int result = subprocess_create(cmdl,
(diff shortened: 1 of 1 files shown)


dev/src/platform.cpp
```diff
@@ -568,32 +568,36 @@ void SetConsole(bool on) {
     #endif
 }
 
-iint LaunchSubProcess(const char **cmdl, const char *stdins, string &out) {
+iint LaunchSubProcess(const char** cmdl, const char* stdins, string& out) {
```
aardappel, line 571: please undo, here and elsewhere

dev/src/platform.cpp
```diff
@@ -568,32 +568,36 @@ void SetConsole(bool on) {
...
         if (stdins) {
-            FILE *p_stdin = subprocess_stdin(&subprocess);
+            FILE* p_stdin = subprocess_stdin(&subprocess);
             fputs(stdins, p_stdin);
             fclose(p_stdin);
         }
+        const unsigned int buflen = 256;
+        const char buf[buflen] = "";
+        unsigned int readlength = subprocess_read_stdout(&subprocess, (char* const)buf, buflen);
+        while (readlength) {
```
aardappel, line 586: why have same code twice? refactor into `for(;;) { .. if (..) break; .. }` loop.

aardappel: the old code had a lambda to do the same 2 identical loops, why not retain that?

AntonBogun: I am not entirely sure how to make a lambda in this case - I would have to pass in the functions themselves as an argument?

dev/src/platform.cpp
AntonBogun: Do I need to change (char* const) to (char *const) as well?

dev/src/platform.cpp
AntonBogun: Should I use std::bind?

dev/src/platform.cpp
aardappel: yes

dev/src/platform.cpp
aardappel: Just use the same code as was there originally?

aardappel: look at the code you replaced.

AntonBogun: So it would look something like `unsigned int (*subprocess_read)(subprocess_s *const, char *const, const unsigned int)`

aardappel:
Nice!

PR #189, opened 2022-07-22
Collection of small fixes and improvements

bartwe:
Fix substring range check if start is beyond the end of the string but a negative size is to copy the remainder.
Make string_to_float work like string_to_int
Add quaternion normalize and mix
Add float4_x/y/z/w constants
Add im_disabled for disabling imgui components


dev/src/lobster/geom.h
```diff
@@ -574,6 +578,13 @@ struct quat : float4 {
     }
 };
 
+inline quat normalize(const quat &q) { return quat(normalize((float4)q)); }
+
+inline quat mix(quat a, quat b, float f) {
+    if(dot(a, b) < 0.0f) b.flip();
+    return normalize(quat(mix((float4)a, (float4)b, f)));
```
aardappel, line 585: Not sure what this is going to be used for, but if its character animation, a linear interpolation is likely not great? How about a spherical lerp instead?

This from some old character animation lib of mine:

```
struct quat : vec4
{    
    quat() {}
    
    void restorew()    { w = 1.0f-x*x-y*y-z*z; w = w<0 ? 0 : sqrtf(w); }
    
    void qmul(quat &o) { quat p = *this;    
                         w = p.w*o.w - p.x*o.x - p.y*o.y - p.z*o.z;
                         x = p.w*o.x + p.x*o.w + p.y*o.z - p.z*o.y;
                         y = p.w*o.y - p.x*o.z + p.y*o.w + p.z*o.x;
                         z = p.w*o.z + p.x*o.y - p.y*o.x + p.z*o.w; }
    
    float dot(quat &o) { return vec::dot(o)+w*o.w; }
    float squaredlen() { return dot(*this); }
    float magnitude()  { return (float)sqrtf(squaredlen()); }

    void normalize()   { float s = squaredlen(); if(s>0) mul(1/sqrtf(s)); }

    void slerp(quat &from, quat &to, float t)
    {
        *this = from;
        quat to1 = to;
        float cosom = from.dot(to);
        if(cosom<0) { cosom = -cosom; to1.neg(); }
        
        if((1-cosom)>1e-6) 
		{
            float omega = acosf(cosom);
            float sinom = sinf(omega);
            mul(sinf((1-t)*omega)/sinom);
            t = sinf(t*omega)/sinom;
        }
        else
        {
            mul(1-t);
        }

        to1.mul(t);
        add(to1);
    }
};
```

dev/src/imbind.cpp
```diff
@@ -705,6 +705,21 @@ nfr("im_group_end", "", "", "",
         NPop(vm, N_GROUP);
     });
 
+nfr("im_disabled_start", "label", "B", "B",
```
aardappel, line 708: `label` is probably not intended

PR #190, opened 2022-07-22
Bartwe/named voxels

bartwe:
Add support for named voxel model in magikavoxel .vox files


dev/src/builtins.cpp
```diff
@@ -375,7 +375,7 @@ nfr("substring", "s,start,size", "SII", "S",
         iint size = e.ival();
         iint start = s.ival();
         if (size < 0) size = l.sval()->len - start;
-        if (start < 0 || start + size > l.sval()->len)
```
aardappel, line 378: would be good to rebase so these other changes are not in here

dev/src/cubegen.cpp
```diff
@@ -563,6 +569,12 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
             auto p = buf.c_str() + 8;
             bool chunks_skipped = false;
             Voxels *voxels = nullptr;
+            map<int32_t, int32_t> node_graph;
```
aardappel, line 572: map is an expensive data structure, are the indices so sparse that its necessary (over `vector`)?

dev/src/lobster/tools.h
```diff
@@ -1037,6 +1037,22 @@ template<typename T> T ReadMemInc(const uint8_t *&p) {
     return dest;
 }
 
+template<typename T> void ReadVec(const uint8_t *&p, T &v) {
```
aardappel, line 1040: Might as well call this `ReadVec64` since 64-bit serialized lengths are not exactly common in the wild.
In fact, if this function is not used lets leave it out all-together. Or maybe better, just make the index type a template argument.

dev/src/lobster/cubegen.h
```diff
@@ -32,6 +32,7 @@ struct Voxels {
     bool chunks_skipped = false;
     Chunk3DGrid<uint8_t> grid;
     int idx = 0;
+    string name = "";
```
aardappel, line 35: no need for the `= ""`, that is the default.

dev/src/cubegen.cpp
```diff
@@ -589,15 +601,102 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
...
+                    }
                 } else {
                     chunks_skipped = true;
                 }
                 p += contentlen;
             }
+            for (auto &i : node_to_layer)
+                if (layer_names.contains(i.second) && !node_names.contains(i.first))
+                    node_names.insert_or_assign(i.first, layer_names[i.second]);
+            for (auto &i : node_to_model) {
```
aardappel, line 687: So all those maps are needed to just generate a single string? I mean, it's nice to have this code in the sense that we may later want to do more with this hierarchy data, but I would not want it to slow down `.vox` loading.. a `map` can imply a lot of dynamic allocation, any way you can represent this more efficiently?

Also, how does one set a name in the MV UI? Couldn't figure that last time I checked.

dev/src/cubegen.cpp
bartwe: there is no official documentation that specifies any requirements on the range or sparseness of these ids. a quick look at the file does show small mostly sequential integers.
given the shortlivedness of these structures and the low usage it shouldn't be measurable of a difference ?

dev/src/builtins.cpp
bartwe: rebased

dev/src/lobster/tools.h
bartwe: it is used by octree_bindings and new usages in the skeletal code(though that in term was discussed to go into flatbuffers)

dev/src/lobster/cubegen.h
bartwe: ah, with something auto initializing and others not doing that, what is the pattern to avoid failing to empty initialize when depending on that state ?

dev/src/cubegen.cpp
bartwe: Annoyingly the data is hidden behind model_ids, layer_ids and node_ids who all need to be connected together and all of which can have string->string metadata sections.

Perhaps i should put this resolving behind a setting/bool argument for code that requests this data ?

It is a feature in the latest beta 0.99.7.0

bartwe:
error: no member named 'contains' in 'std::map<int, std::basic_string<char>>' seems like contains is too new, can replace it with find() != end() if prefered

dev/src/cubegen.cpp
bartwe: would unordered_map help a lot ? The datastructure i'm used to use here is backed by two vectors, one is a hashtable, holding an offset into the second vector holding the value avoiding per value allocations. For trivial value types folded down ofc into a single vector

aardappel: `unordered_map` are per lookup if you have a lot of keys, but their fixed overhead may well be even greater than `map`..

We load a LOT of .vox files, so each of those constructing and destructing 5 maps seems excessive. Of course, hard to tell until you measure. If they're small ints I'd vote for at least using `vector`, and I wonder if it can be done even more efficiently than that.

Anyway, you can merge it with `map` if you want to just move on, and note it in a `FIXME` and maybe a future task to optimize if you're bored with other things :)

dev/src/cubegen.cpp
aardappel: yeah a bool could work, or alternatively only parse it if the file contains multiple models, that would also reduce the amount of files for which it needs to happen.

PR #191, opened 2022-07-27
Bartwe/span

bartwe:
Use span for safe bounded access when reading data from buffers that might be too small.


dev/src/lobster/tools.h
```diff
@@ -1031,30 +1031,51 @@ template<typename T> T ReadMem(const void *p) {
...
+    dest += sizeof(T);
 }
 
-template<typename T, typename K = uint64_t> void ReadVec(const uint8_t *&p, T &v) {
-    auto len = ReadMemInc<K>(p);
-    v.resize((size_t)len);
-    auto blen = sizeof(typename T::value_type) * v.size();
-    memcpy(v.data(), p, blen);
-    p += blen;
+template<typename T> bool ReadSpan(const span<const uint8_t> &&p, T &v) {
```
aardappel, line 1039: why the `&&` ?
Spans should probably be passed by value when possible.

dev/src/cubegen.cpp
```diff
@@ -559,134 +559,160 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
...
+                    for (int i = 0; i < dict_len; ++i) {
                         string key;
-                        ReadVec<string, int32_t>(c, key);
-                        if (key == "_name") {
-                            string value;
-                            ReadVec<string, int32_t>(c, value);
-                            layer_names.insert_or_assign(layer_id, value);
-                        } else
-                            SkipVec<string, int32_t>(c);
+                        if (ReadSpanVec<string, int32_t>(p, key))
```
aardappel, line 705: here you're simply skipping data if eof, whereas above its an error, is that intentional?

dev/src/cubegen.cpp
```diff
@@ -559,134 +559,160 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
...
+                            int32_t model_id;
+                            if (ReadSpanInc(p, model_id))
+                                node_to_model.insert_or_assign(node_id, model_id);
+                            int32_t dict_len;
+                            if (ReadSpanInc(p, dict_len))
+                                for (int i = 0; i < dict_len; ++i) {
+                                    SkipSpanVec<string, int32_t>(p);
+                                    SkipSpanVec<string, int32_t>(p);
+                                }
+                            [[maybe_unused]] int32_t reserved;
```
aardappel, line 694: neat use of feature, but it pushes complexity to user of the function.. better to default-init the 2nd arg in the function, and since that only happens on the error path, it should be cheap?

aardappel: alternatively, return a `pair<bool, T>`, and read with `auto [ok, t] =` though I suppose that doesn't allow convenient use inside `if`?

dev/src/cubegen.cpp
```diff
@@ -559,134 +559,160 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
...
+                    if (!ReadSpanInc(p, numvoxels)) return erreof();
+                    if (p.size_bytes() < numvoxels * sizeof(byte4)) return erreof();
+                    auto vp = (uint8_t *)p.data();
                     for (int i = 0; i < numvoxels; i++) {
-                        auto vox = byte4((uint8_t *)(p + i * 4 + 4));
+                        auto vox = byte4((vp + i * 4));
                         auto pos = int3(vox.xyz());
                         if (pos < voxels->grid.dim) voxels->grid.Get(pos) = vox.w;
                     }
+
```
aardappel, line 614: I think I just removed these?

dev/src/lobster/tools.h
bartwe: Ah sorry the c++11 stuff from long ago is leaking, with the const && it'll pass in the argument as an rvalue when it is a temporary.
Not relevant with a simple type like span here. i'll remove it

dev/src/cubegen.cpp
bartwe: i was a bit worried about the cost in a loop, but its a branch anyways so that doesn't make sense in the current code, changed.

dev/src/cubegen.cpp
bartwe: what do you mean ?

dev/src/cubegen.cpp
bartwe: yeah the result is used in if()return and && chainging in a few places in a way symmetric with the FileWrite methods

dev/src/cubegen.cpp
aardappel: you're adding blank lines that I just removed..

PR #193, opened 2022-08-04
Update flatbuffers to 2.0.6

bartwe:


aardappel:
Thanks! Missing some files though, see here last time we upgraded: https://github.com/aardappel/lobster/commit/df371467adbea59a5cf902dfa86f9478f085fb2c

Also.. if you went with 2.0.6 because it is the last tag, that has no special meaning (probably created for a language that is not C++, and older), so you might as well pick latest master instead. That's what all past updates have done.

PR #202, opened 2022-09-19
Add warning about unsupported rotations in a vox file

bartwe:


dev/src/cubegen.cpp
```diff
@@ -672,6 +672,18 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
                                 offset.z = std::strtol(cursor, &next, 10);
                                 node_offset.insert_or_assign(node_id, offset);
                             }
+                            if (key == "_r") {
+                                const char* cursor = value.c_str();
```
aardappel, line 676: `*` always next to name

dev/src/cubegen.cpp
```diff
@@ -672,6 +672,18 @@ nfr("cg_load_vox", "name", "S", "R:voxels]S?",
...
+                                char* next;
+                                auto rotation = std::strtol(cursor, &next, 10);
+                                // 4 is the noop rotation
+                                if (rotation != 4) {
+                                    string msg = ".vox file uses an object rotation that is not supported. :";
+                                    msg += value;
+                                    printf("%s\n", msg.c_str());
```
aardappel, line 683: we don't want `printf` in any "library" code, use `errf`

aardappel (commented):
Good idea!

aardappel (approved):

PR #203, opened 2022-09-20
Improve file loading error messages

bartwe:


aardappel (commented):
Thanks, that's useful.. will have to see if this produces any log-spam, because under some circumstances these functions may be called just to check if a file is there, in which case it should not produce output (it is not an error). If that happens, we'll have to move these messages to the caller. But for now we can merge.

Issue #205, opened 2022-09-21
music.c: Fails to compile with empty union.


dbaileychess:
Grabbed the latest source and try to build on Windows (VS 2022, Release X64) and it fails to compile with:

```
lobster\dev\external\SDLMixer\music.c(120,5): error C2016: C requires that a struct or union have at least one member
```

Appears I don't have any of [these](https://github.com/aardappel/lobster/blob/master/dev/external/SDLMixer/music.c#L84-L120) defined that leads to the empty union. I can probably get the necessary files, but either the Docs should specify how, or something is not included correctly by default.

Steps:

```
git clone 
cd dev
cmake -DCMAKE_BUILD_TYPE=Release
```

Opened  `dev/Lobster.sln` in VS2022, switched to `Release` and `Rebuilt Solution`

aardappel:
Don't use CMake on Windows.. there's an already made VS solution in dev/lobster that has lots of custom flags set that aren't in CMake (never tried building with that on Windows).

See also: https://aardappel.github.io/lobster/getting_started.html

PR #206, opened 2022-09-22
Disable CMake on non-linux systems

dbaileychess:
CMake generation is only supported for Linux/BSD at the moment, so prevent CMake generation on non-supportive platforms.

This would have helped make #205 more obvious.

### DIFF
--- a/dev/CMakeLists.txt
+++ b/dev/CMakeLists.txt
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 2.8)
 
+if(NOT UNIX) 
+  message(FATAL_ERROR "Cmake configuration is only supported for Linux. For "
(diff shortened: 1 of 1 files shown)


aardappel:
Mind adding, "for Windows/Mac/iOS/Android/Emscripten, use the included native projects instead" or similar verbiage?

deathstar:
I thought cmake worked on all os’ but you just needed to use a GUI on windows.

aardappel:
@deathstar cmake itself works on Windows even without a gui.. but the cmake script in this case doesn't set all the right options that the VS project does. This could be ported over but is a bit of work.

dbaileychess:
Done.

aardappel:
Thanks!

PR #207, opened 2022-09-24
Support Scalar OP Numeric Structs

dbaileychess:
Currently, the binary ops for `numeric struct` and a `scalar` only work if the left value is the `Numeric Struct`:

```
xy_i{1,5} * 4 
```

This PR makes it so the scalar can be the left value.

```
4 * xy_i{1,5}
```

See `tests/structtest.lobster` for all the implications.


aardappel:
Wow, this is done with less changes than I expected, and such a nice improvement!

One thing is that you've chose not to reuse the vector ops that are commutative, which leads to more code, but thinking about I may prefer it this way, since it keeps evaluation order. E.g. `f() * g()` where these functions have side effects would need some kind of argument swap operation to be passed to `*` when reusing the commutative vector op, OR I would need to declare that the evaluation order for operators is implementation defined, whereas so far we have a nice left to right order.

PR #209, opened 2022-09-25
Add `matrix.lobster` for 2x2 float matrices

dbaileychess:
Makes a new lobster library for handling 2x2 matrices by expressing them in terms of `xyzw_f`. This way a lot of the operators are already well-defined, and only the multiplicative ones need special handling.

### DIFF
--- a/modules/matrix.lobster
+++ b/modules/matrix.lobster
@@ -7,16 +7,12 @@ import vec
 struct mat2x2 : xyzw_f
 
     def operator*(o:mat2x2):
-        // This could be expressed with 4 dot product if `this.zw` existed (and 
-        // others)
-        return mat2x2 { x*o.x + y*o.z, x*o.y + y*o.w,
-                        z*o.x + w*o.z, z*o.y + w*o.w }  
+        return mat2x2 { dot(this.xy, o.xz), dot(this.xy, o.yw),
+                        dot(this.zw, o.xz), dot(this.zw, o.yw) }
     
     def operator*(o:xy_f):
-        // This would be nicer if `this.zw` existed.
         return xy_f { dot(o, this.xy),
-                      dot(o, xy_f{z,w})}
-
+                      dot(o, this.zw) }
 
     def operator*(f:float):
         // Note the order is flipped so that it is called with xyzw_f * operator
@@ -39,4 +35,4 @@ struct mat2x2 : xyzw_f
 // The identity matrix:
 // |1 0|
 // |0 1|
-let mat2x2_identity = mat2x2{1.0, 0.0, 0.0, 1.0}
+let mat2x2_identity = mat2x2 { 1.0, 0.0, 0.0, 1.0 }
--- a/modules/matrix.lobster
+++ b/modules/matrix.lobster
@@ -1,0 +1,42 @@
+import vec
+
+// A 2x2 matrix representated as a 4 length vector of x,y,z,w so that
+// |x y|
+// |z w|
+
+struct mat2x2 : xyzw_f
+
+    def operator*(o:mat2x2):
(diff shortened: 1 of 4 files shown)


modules/matrix.lobster
```diff
@@ -0,0 +1,42 @@
...
+struct mat2x2 : xyzw_f
+
+    def operator*(o:mat2x2):
+        // This could be expressed with 4 dot product if `this.zw` existed (and 
+        // others)
+        return mat2x2 { x*o.x + y*o.z, x*o.y + y*o.w,
+                        z*o.x + w*o.z, z*o.y + w*o.w }  
+    
+    def operator*(o:xy_f):
+        // This would be nicer if `this.zw` existed.
```
aardappel, line 16: just add `zw` in `vec.lobster`

modules/matrix.lobster
```diff
@@ -0,0 +1,42 @@
...
+        return adjugate() / determinant()
+
+    def trace():
+        return x*w
+
+        
+// The identity matrix:
+// |1 0|
+// |0 1|
+let mat2x2_identity = mat2x2{1.0, 0.0, 0.0, 1.0}
```
aardappel, line 42: spaces around `{` and `}`

modules/matrix.lobster
```diff
@@ -0,0 +1,42 @@
...
+        return mat2x2 { x*o.x + y*o.z, x*o.y + y*o.w,
+                        z*o.x + w*o.z, z*o.y + w*o.w }  
+    
+    def operator*(o:xy_f):
+        // This would be nicer if `this.zw` existed.
+        return xy_f { dot(o, this.xy),
+                      dot(o, xy_f{z,w})}
+
+
+    def operator*(f:float):
```
aardappel, line 21: we could leave out this operator entirely?

tests/unittest.lobster
```diff
@@ -19,6 +19,7 @@ import gradienttest
 import springstest
 import smallpttest
 import stringtest
+import matrixtest
```
aardappel, line 22: add it lower since its not in speedtest

tests/matrixtest.lobster
```diff
@@ -0,0 +1,42 @@
...
+
+    do():
+        let a = mat2x2 { 1.0, 2.0, 3.0, 4.0 }
+        let b = mat2x2 { 5.0, 6.0, 7.0, 8.0 }
+
+        assert a == a
+        assert a == mat2x2 { 1.0, 2.0, 3.0, 4.0 }
+        assert a != b
+
+        assert 5.0 + a == mat2x2{ 6.0, 7.0, 8.0, 9.0 }
```
aardappel, line 14: same here space before `{`

modules/matrix.lobster
dbaileychess: I was worried that it could be confusing, since we are shifting the z -> x and w-> y, but if you are OK with it.

modules/matrix.lobster
dbaileychess: Sorry, the highlighting is odd. Which operator, the *float?

aardappel: yes this entire function, `def operator*(f:float)`

modules/matrix.lobster
aardappel: that's the same as in GLSL, if you do `.zw` you'll get an xy vector back

modules/matrix.lobster
dbaileychess: Ok, removed. Those it does remove this `assert a * 5.0 == mat2x2 { 5.0, 10.0, 15.0, 20.0 }` from working.

aardappel:
Thanks!

PR #211, opened 2022-09-30
Updated SublimeText Syntax

Nieb:
Removed XML lobster.tmLanaguge file, created and added YAML lobster.sublime-syntax.
It is usable, but still a work-in-progress.


aardappel:
Thanks!

Issue #216, opened 2022-11-14
Assertion failed on size for small_vector<>


stolk:
Building with gcc-11 on Ubuntu:

```
$ ./lobster 
lobster: /home/bram/src/lobster/dev/src/lobster/tools.h:1230: void unit_test_tools(): Assertion `sizeof(small_vector<int, 2>) == 12' failed.
Aborted (core dumped)
```

```
$ cmake -DCMAKE_BUILD_TYPE=Debug ../dev/
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- The C compiler identification is GNU 11.3.0
-- The CXX compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Deprecation Warning at external/SDL/CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 
-- Performing Test HAVE_GCC_WALL
-- Performing Test HAVE_GCC_WALL - Success
-- Performing Test HAVE_GCC_NO_STRICT_ALIASING
-- Performing Test HAVE_GCC_NO_STRICT_ALIASING - Success
-- Performing Test HAVE_GCC_WDECLARATION_AFTER_STATEMENT
-- Performing Test HAVE_GCC_WDECLARATION_AFTER_STATEMENT - Success
-- Performing Test HAVE_GCC_WERROR_DECLARATION_AFTER_STATEMENT
-- Performing Test HAVE_GCC_WERROR_DECLARATION_AFTER_STATEMENT - Success
-- Performing Test HAVE_GCC_ATOMICS
[...]

PR #217, opened 2022-11-14
Struct packing rules should result to a size of 16.

stolk:
So, we really expect the struct size to be 16 here.

To be safe, I used `<= 16` in case Visual C++ does something strange.

*For struct, other than the alignment need for each individual member, the size of whole struct itself will be aligned to a size divisible by strictest alignment requirement of any of its members, by padding at end.*

Fixes issue #216


aardappel:
Apologies I pre-empted you.

Decided that if most compilers were going to add padding, might as well use that padding for larger fields, and get rid of all the (apparently not portable) packing trickery: https://github.com/aardappel/lobster/commit/3b2deb5289b87a8f1a67ef01ac0c448549b558b9

PR #223, opened 2023-01-30
Texture initialization rework

maierfelix:
`gl_create_blank_texture` now creates a texture without initializing the data, while `gl_create_colored_texture` is used for initializing a texture with a given color


aardappel:
Thanks!

PR #225, opened 2023-02-13
Add offset parameter to gl_update_buffer_object

maierfelix:


aardappel:
Please rebase, this contains changes from other commits

dev/src/cubegen.cpp
```diff
@@ -779,14 +779,7 @@ nfr("cg_load_vox", "name,material_palette", "SI?", "R:voxels]S?",
...
                     }
-                    // FIXME: for the files that have one of these chunks in,
-                    // the palette entries only correspond correctly to the voxels if
-                    // you DON'T apply this remapping.
-                    // So what is it for?
-                    // The only more extensive description is here and it doesn't seem
-                    // to be correct:
-                    // https://github.com/ephtracy/voxel-model/issues/19
-                    //palette = remapped_palette;
+                    palette = remapped_palette;
```
aardappel, line 782: why is this reinstated?

maierfelix: Weird I think it came with the rebase, don't think it was there before

PR #238, opened 2023-04-17
Allow --import <dir> to be absolute

ewoudje:
Checks if the given path is a absolute path if so don't add project path in front.

### DIFF
--- a/dev/src/platform.cpp
+++ b/dev/src/platform.cpp
@@ -158,6 +158,11 @@ string_view StripTrailing(string_view in, string_view tail) {
     return in;
 }
 
+
+bool IsAbsolute(string_view filename) {
+    return filename.size() > 1 && (filename[0] == '/' || filename[0] == '\\' || filename[1] == ':');
+}
+
 string GetMainDirFromExePath(string_view argv_0) {
     string md = SanitizePath(argv_0);
     #ifdef _WIN32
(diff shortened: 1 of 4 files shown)


dev/src/platform.cpp
```diff
@@ -279,9 +284,10 @@ bool InitPlatform(string _maindir, string_view auxfilepath, bool from_bundle,
     return true;
 }
 
-void AddDataDir(string_view path, bool is_rel) {
-    auto fpath = SanitizePath(path);
-    if (is_rel) fpath = projectdir + fpath;
+void AddDataDir(string_view path) {
+    // If the import is written like --import /project/abc add slash at the end.
```
aardappel, line 288: shouldn't you check if there already is a `FILESEP` at the end?

ewoudje: If I remember correctly, its intentional design in file paths that double slashes have no meaning (Atlough I haven't seen it on black and white so take it with a grain of salt).
So D:\Project\\test.lobster is valid and for linux /project//test.lobster is also valid.

aardappel: I know, but just because it is legal doesn't mean we don't want to generate nice looking paths :)

aardappel:
thanks!

aardappel:
`ends_with` : for some reason, MS is further ahead with implementing the latest STL features than gcc/clang..

PR #239, opened 2023-04-18
The LSP and the VSCode addon using it

ewoudje:
The LSP implementation and the vscode extension that i have showcased before.
Its WIP still; but the first wave of recommendations/changes can be made already.

Both parts are written in typescript.

TODO:
* [x]  Add experimental option and put `--errors` under it.
* [x]  Conflict when parsing and writing the file at the same time.
* [x]  Github workflow for testing lsp
* [x] Fix code needing manual modification before packaging
* [x] Update main docs on how to use the extension and its features
* [x] Add lobster icons
* [x]  Add breakpoints via DAP (Finished kinda has debugging but no breakpoints, not gonna do this in this PR)


dev/lsp/src/document.ts
```diff
@@ -0,0 +1,305 @@
...
+	public update(changes: TextDocumentContentChangeEvent[], version: number): void {
+		for (const change of changes) {
+			if (LobsterDocument.isIncremental(change)) {
+				// makes sure start is before end
+				const range = getWellformedRange(change.range);
+
+				// update content
+				const startOffset = this.offsetAt(range.start);
+				const endOffset = this.offsetAt(range.end);
+				this._content = this._content.substring(0, startOffset) + change.text + this._content.substring(endOffset, this._content.length);
```
aardappel, line 142: user a shorter line length so source is easier to view/diff etc?

dev/lsp/tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
...202,105],[204,105],[213,98],[42,98],[43,98],[9,98],[8,98],[2,98],[10,98],[11,98],[12,98],[13,98],[14,98],[15,98],[16,98],[17,98],[3,98],[4,98],[21,98],[18,98],[19,98],[20,98],[22,98],[23,98],[24,98],[5,98],[25,98],[26,98],[27,98],[28,98],[6,98],[32,98],[29,98],[30,98],[31,98],[33,98],[7,98],[34,98],[39,98],[40,98],[35,98],[36,98],[37,98],[38,98],[1,98],[41,98]],"semanticDiagnosticsPerFile":[107,108,110,111,112,113,114,115,116,117,118,119,120,122,121,123,124,125,109,159,126,127,128,160,129,130,131,132,133,134,135,136,137,138,139,140,141,143,142,144,145,146,147,148,149,150,151,152,153,154,155,156,157,158,168,58,50,55,48,51,49,47,57,53,46,54,52,56,162,163,45,82,81,59,69,64,63,80,66,78,73,65,60,77,76,72,74,79,68,67,70,71,61,75,62,164,165,100,44,99,86,84,94,89,93,92,90,97,96,83,87,98,88,95,91,85,161,166,167,106,103,101,102,170,171,104,169,105,172,197,198,173,176,195,196,186,185,183,178,191,189,193,177,190,194,179,180,192,174,181,182,184,188,199,187,175,212,211,206,208,207,200,201,203,205,209,210,202,204,213,42,43,9,8,2,10,11,12,13,14,15,16,17,3,4,21,18,19,20,22,23,24,5,25,26,27,28,6,32,29,30,31,33,7,34,39,40,35,36,37,38,1,41],"latestChangedDtsFile":"./out/utils.d.ts"},"version":"5.0.4"}
```
aardappel, line 1: I am not familiar with what files are typically checked into a repo in TS.. this file seems very much generated.. can you double check for all files?

dev/src/lobster/parser.h
```diff
@@ -171,7 +171,7 @@ struct Parser {
                 if (IsNext(T_FROM)) {
                     string fn = std::move(lex.sval);
                     Expect(T_STR);
-                    AddDataDir(fn, true);
```
aardappel, line 174: You need to rebase

docs/vscode/lobster/.eslintrc.js
```diff
@@ -0,0 +1,20 @@
+/**@type {import('eslint').Linter.Config} */
+// eslint-disable-next-line no-undef
+module.exports = {
```
aardappel, line 3: why is this stuff in this dir and not in LSP? same for the files below

docs/vscode/lobster/.vscode/launch.json
```diff
@@ -8,10 +8,15 @@
...
-            "args": [
-                "--extensionDevelopmentPath=${workspaceFolder}"
-            ]
+			"request": "launch",
+			"name": "Launch Client",
+			"runtimeExecutable": "${execPath}",
+			"args": ["--extensionDevelopmentPath=${workspaceRoot}"],
+			"outFiles": ["${workspaceRoot}/client/out/**/*.js"],
+			"preLaunchTask": {
+				"type": "npm",
```
aardappel, line 17: is this correct? should users of a plugin run `npm` all the time?

docs/vscode/lobster/package-lock.json
```diff
@@ -0,0 +1,2843 @@
+{
+    "name": "lobster",
```
aardappel, line 2: users of a plugin should not need all this stuff

docs/vscode/lobster/.vscode/launch.json
ewoudje: This is when you are testing the plugin, it runs `npm watch` so to hot update typescript to js.

aardappel: Right, but stuff in this dir is what users copy to their `.vscode` folders. If this stuff is only used during development, then we should have a folder inside `dev/lsp` that holds this stuff, and keep this folder clean for users.

docs/vscode/lobster/.eslintrc.js
ewoudje: I forgot, fixed

docs/vscode/lobster/.vscode/launch.json
ewoudje: Hmm, this is true normally .vscodeignore is used for this. When you publish a extension those files are not included. I can move the actual implementation away but im unsure what to put here then.

aardappel: A `.vscodeignore` is not used by windows file copy afaik ;) Users will end up with lots of dev files they don't use.

You don't need anything here. This is a folder users copy. You can make a similar folder in `dev/lsp` that contains all these files, and work/test from there instead of this folder.

ewoudje: But the extension is based of the files in (lets call it) `dev/vscode`. Normaly there runs a process that uses .vscodeignore for this. We can automate this with github actions but then its a commit that updates this after every commit. And manualy updating it sounds also a bad idea. Another solution would remove this folder and have a artifact be prepared.

aardappel: We can either do it manually, or in CI, or maybe better yet, this should be uploaded to the VSCode extensions repository so we can just tell users to get the extension inside VSCode?

I suggest to start with manual for now, then find out better ways later. It is not that much work to run whatever command packages the extension before you commit any changes to the code. Just add a .bat/.sh

Regardless, user and dev files should not be mixed.

docs/vscode/tasks.json
```diff
@@ -1,42 +1,17 @@
...
-                    "file": 1,
-                    "line": 2,
-                    "severity": 3,
-                    "message": 4
-                }
-            }
+			// A example of a lobster task
+            "label": "Run lobster",
+            "type": "lobster",
+            "main": "main.lobster",
```
aardappel, line 11: this doesn't seem correct.. the above code you deleted can run the current file, this would only be able to run `main.lobster` ? Does the extension take care of platform differences?

`tasks.json` is a file people can copy to their project, as opposed to being an extension.. it should probably remain working even if the lobster extension is not installed

aardappel:
Need to update the main docs on how to install the extension.

Also, you moved the syntax highlighting file into the extension. Is this normal? I would expect that to function regardless of whether you are using an LSP.

ewoudje:
> Need to update the main docs on how to install the extension.
> 
> Also, you moved the syntax highlighting file into the extension. Is this normal? I would expect that to function regardless of whether you are using an LSP.

Well it was a very small extension, now I just made it a bigger extension that boots up a LSP.

ewoudje:
With the latest commit it now includes a packaged file with the full extension which is very easy to install.

docs/vscode/tasks.json
ewoudje: Oh ok, i thought it was meant as a example on how to use the extension, Il revert that back

ewoudje:
Im unsure if I should modify some READMEs to highlight the existence of a vscode extension?

docs/source/lsp.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Lobster LSP
+---
+
+TODO
```
aardappel, line 5: Yes, lets have some description on how to install it and use it :)

docs/source/vscode_ide.md
```diff
@@ -2,19 +2,25 @@
 title: VSCode as Lobster IDE
 ---
 
-Running Lobster programs from within the editor
+VSCode extension
+-------------------
+The vscode extension includes syntax highlithing,
```
aardappel, line 7: `highlighting`

docs/source/vscode_ide.md
```diff
@@ -2,19 +2,25 @@
 title: VSCode as Lobster IDE
 ---
 
-Running Lobster programs from within the editor
+VSCode extension
+-------------------
+The vscode extension includes syntax highlithing, 
+a lobster lsp and breakpoints.
```
aardappel, line 8: `LSP (Language Server Protocol)`

docs/source/vscode_ide.md
```diff
@@ -2,19 +2,25 @@
...
+
+See [here](command_line_usage.html) what the lsp is capable of.
```
aardappel, line 10: `LSP`

docs/source/vscode_ide.md
```diff
@@ -2,19 +2,25 @@
...
+
+### How to install
+-  Download the file `lobster.vsix` found in the github at `docs/vscode/lobster.vsix`.
```
aardappel, line 13: `the github` -> `the repo`

docs/source/vscode_ide.md
```diff
@@ -2,19 +2,25 @@
...
+
+-  Open your vscode command pallete and search for `Install from VSIX`. Then select the file you just downloaded.
```
aardappel, line 15: It's not really "downloading" if most people reading this will already have the Lobster repo

aardappel (commented):
Generally looks great! A bit more work on the docs and we can merge this :)

docs/source/lsp.md
ewoudje: Oops forgot that I left a TODO there 😅

aardappel:
Awesome!

PR #240, opened 2023-04-18
Add favicon to docs

binji:
Looks like this:

msedge_mZ9DUynuqx

Just using the existing lobster.svg as an icon.


aardappel:
Neato!

PR #243, opened 2023-04-24
Add cg_load_vox support for rotation/duplication

binji:


dev/src/cubegen.cpp
```diff
@@ -821,7 +832,27 @@ nfr("cg_load_vox", "name,material_palette", "SI?", "R:voxels]S?",
                 }
             }
             if (voxvec->SLen() < (ssize_t)node_to_model.size()) {
-                return errf(".vox file uses object deduplication feature that is not supported\n");
+                auto new_voxvec = vm.NewVec(0, 0, TYPE_ELEM_VECTOR_OF_RESOURCE);
```
aardappel, line 835: doesn't matter too much, but more elegant to just call `voxvec->Push` which will enlarge the existing vector?

dev/src/lobster/3dgrid.h
```diff
@@ -69,6 +70,24 @@ template<typename T> class Chunk3DGrid : NonCopyable {
         }
         dim = ndim;
     }
+
+    void Rotate(const matrix<int, 3, 3> &m) {
```
aardappel, line 74: I guess this is different in setup from what goes on in `cg_rotate` ? Can you check? Fine to keep this copy if its messy to share.

binji: Yeah, it's different in that `cg_rotate` only rotates around Z by 90 degrees. We could probably make that one use this logic, though

dev/src/cubegen.cpp
binji: I think the problem may be that the new voxels aren't necessarily at the end, so I'd need to insert into the middle of the vector

aardappel: There's an `Insert` too :)

binji: Oh heh, right! Well, I guess theoretically the vox file could even have the shape indexes out of order too, but I doubt that ever happens in practice. I think this code gets a bit clumsier regardless, since you would need to copy from the vec as it's being modified

PR #282, opened 2023-12-11
Try to fix emscripten build

binji:


aardappel:
Wow you fixed the Makefile, that's heroic :)

PR #286, opened 2023-12-30
Allow toggling comments in Sublime Text

oconnor0:
Allows Ctrl+/ to toggle line comments.
Allows Ctrl+Shift+/ to toggle block comments.


aardappel:
Nice :)

PR #290, opened 2024-01-11
add more window init flags

Jack-Ji:
Add flags for controlling resizability and border, which should be useful when developing small/retro games.


aardappel:
LGTM

PR #293, opened 2024-01-23
Add missing methods for dictionary

Jack-Ji:


modules/dictionary.lobster
```diff
@@ -35,15 +36,37 @@ class dictionary<K,V>:
...
+            c = c.next
+        guard c
+        sz--
+        if c == buckets[h]:
+            buckets[h] = c.next
+        else:
+            guard prev
+            prev.next = c.next
+
+    def size():
```
aardappel, line 59: what is this member/function useful for? typically a size is only useful in combination with iterating all elements, for which we so far only have `get_all`

modules/dictionary.lobster
```diff
@@ -35,15 +36,37 @@ class dictionary<K,V>:
...
+            if c.key == key:
+                break
+            prev = c
+            c = c.next
+        guard c
+        sz--
+        if c == buckets[h]:
+            buckets[h] = c.next
+        else:
+            guard prev
```
aardappel, line 56: as much as I love `guard`, this here is probably more readfable as `elif prev:`

modules/dictionary.lobster
```diff
@@ -35,15 +36,37 @@ class dictionary<K,V>:
...
             c.value = value
         else:
+            sz++
             buckets[h] = chain<K,V> { buckets[h], key, value }
 
+    def del(key:K):
+        let h = hash(key) % buckets.length
+        var c = buckets[h]
+        var prev = nil::chain<K,V>
+        while c:
```
aardappel, line 46: this algorithm is a bit lengthy and clumsy, probably would be neater as:
```
def delete(c):
    if c.key == key:
        return c.next
    c.next = delete(c.next)
    return c
buckets[h] = delete(buckets[h)
```

modules/dictionary.lobster
Jack-Ji: Initially I thought it would be convenient to use like std::set, e.g. for counting unique records. On second thought, keys will be eventually checked after that. So yeah, maybe it's a redundant method indeed.

modules/dictionary.lobster
Jack-Ji: It is neater, at the cost of more overhead with recursion calls. Though hash collision should be considered rare case I guess? Will recommit later

modules/dictionary.lobster
```diff
@@ -69,5 +83,29 @@ class dictionary<K,V>:
...
+    def get_values() -> [V]:
+        let vals = []
+        for(buckets) c:
+            while c:
+                vals.push(c.value)
+                c = c.next
+        return vals
+
+    def forkv(fun):
+        let keys = []
```
aardappel, line 103: these two variables are unused?

modules/dictionary.lobster
```diff
@@ -69,5 +83,29 @@ class dictionary<K,V>:
                 c = c.next
         return keys, vals
 
+    def get_keys() -> [K]:
```
aardappel, line 86: If we're going to have these two methods, lets delete `get_all()`

modules/dictionary.lobster
Jack-Ji: oops

modules/dictionary.lobster
Jack-Ji: ok, will do

PR #295, opened 2024-02-02
Fix resource leak in samples/dragtest.lobster

Jack-Ji:
In old implementation, `lines` will keep accumulating.


aardappel:
Thanks.. I changed the event handling and never updated this..

PR #296, opened 2024-02-04
Add builtin methods for querying mesh animation infos

Jack-Ji:


dev/src/glloadiqm.cpp
```diff
@@ -303,6 +305,12 @@ bool loadiqmanims(const iqmheader &hdr, const char *buf) {
             frames[i*hdr.num_poses + j] = m;
         }
     }
+    const char *str = hdr.ofs_text ? &buf[hdr.ofs_text] : "";
+    anim_names = new const char *[numanims];
+    memset(anim_names, 0, numanims*sizeof(const char *));
```
aardappel, line 310: space around `*`

samples/dragtest.lobster
```diff
@@ -20,10 +20,10 @@ while gl.frame():
     let start = float(gl.last_pos("mouse1", true))
     let cur = float(gl.mouse_pos(0))
     if magnitude(start - cur) > 1.0:  // a click is not a drag
-        let mouse1 = gl.button("mouse1")
```
aardappel, line 23: Rebase

dev/src/graphics.cpp
```diff
@@ -930,6 +930,28 @@ nfr("mesh_size", "m", "R:mesh", "I",
...
+        for (const auto &it : m.animations) v->Push(vm, Value(vm.NewString(it.first)));
+        return Value(v);
+    });
+
+nfr("mesh_animation_frames", "m,name", "R:meshS", "I]?",
+    "given name, return animation's frames range [first_frame, num_frames], or nil if name is invalid",
+    [](StackPtr &, VM &vm, Value &i, Value &n, Value &f) {
+        auto &m = GetMesh(i);
+        if (auto it = m.animations.find(string(n.sval()->strv())); it != m.animations.cend()) {
+            auto v = (LVector *)vm.NewVec(0, 2, TYPE_ELEM_VECTOR_OF_INT);
```
aardappel, line 947: Just use two return values instead of allocating a vector? Make the values -1, -1 if not found.

dev/src/lobster/glinterface.h
```diff
@@ -207,6 +207,11 @@ struct Mesh : lobster::Resource {
     float3x4 *mats = nullptr;
     float curanim = 0;
 
+    struct Animation {
+        uint32_t first_frame, num_frames;
```
aardappel, line 211: Just use `int`.. all Lobster values are signed, so less unsigned values means less conversions on the C++ side.

dev/src/glloadiqm.cpp
```diff
@@ -155,6 +155,7 @@ static const char **textures = nullptr;
 static iqmjoint *joints = nullptr;
 static iqmpose *poses = nullptr;
 static iqmanim *anims = nullptr;
+static const char **anim_names = nullptr;
```
aardappel, line 158: Not your problem, but one day we should stick all these globals in a class, hah!

aardappel (commented):
Looks great!

dev/src/graphics.cpp
Jack-Ji: Multiple return values does make more sense, since I've being thinking how to properly return frame range and framerate, which is float number.

Jack-Ji:
I changed return type of gl.mesh_animation_frames, please review again.

aardappel:
lgtm

PR #298, opened 2024-02-05
Add builtin function gl.save_texture

Jack-Ji:
only png format is supported, like screenshot. I think other formats aren't important in most circumstances.


aardappel:
Nice!

PR #299, opened 2024-02-06
Add builtin function to support animation blending

Jack-Ji:
Only support blending 2 animations, which I guess is enough for most situations.


dev/src/gltexture.cpp
```diff
@@ -19,6 +19,7 @@
 #include "lobster/glincludes.h"
 
 #define STB_IMAGE_IMPLEMENTATION
+#define STB_IMAGE_WRITE_IMPLEMENTATION
```
aardappel, line 22: Is this fixing an issue with a previous commit? If so better in a separate PR

dev/src/glgeom.cpp
```diff
@@ -134,14 +134,25 @@ void BindAsSSBO(Shader *sh, string_view_nt name, int id) {
...
+        auto blend_frame = [this](float fr, float3x4 *out) {
+            int frame1 = ffloor(fr);
+            int frame2 = frame1 + 1;
+            float frameoffset = fr - frame1;
+            float3x4 *mat1 = &mats[(frame1 % numframes) * numbones],
+                     *mat2 = &mats[(frame2 % numframes) * numbones];
+            for(int i = 0; i < numbones; i++) out[i] = mix(mat1[i], mat2[i], frameoffset);
+        };
+
+        auto outframe = new float3x4[numbones*2];
```
aardappel, line 147: Spaces around operators.

Using unused memory in the base case doesn't feel super clean to me, but I also understand its at virtually no cost since its temp.. so I guess its ok.

aardappel (commented):
This generally looks good, as I was expecting this to look :)

Jack-Ji:
About modifications to sdlsystem.cpp/gltexture.cpp, I was doing some minor refactor when implementing animation blending, force of old habit. I'll open a separate PR for it later. :-)

aardappel:
Thanks!

PR #301, opened 2024-02-14
docs: Fix typos

jon-edward:
Tried to be as faithful as possible to the original documentation! Cleaned up a few typos.


jon-edward:
Tried to rebuild HTML docs using `generate_html_docs_from_markdown.sh` and it made an unexpected number of line changes...

The updated docs on my side look visually similar to the docs hosted at http://aardappel.github.io/lobster, not sure what's causing the difference.

aardappel:
what's your `pandoc --version` ? I'm on `3.1.2`

aardappel:
Ok, guess my pandoc was pretty old, so I upgraded to 3.1.11.1 (latest release on the pandoc site) which makes some of the changes that yours also makes. I've committed those for the whole repo.

It does however not reformat all parapgraphs to be wider, which causes most of the churn in your diff. So:

1. Remove all the .html files from your PR.
2. Check your version.. if behind 3.1.11.1, see if you can update.
3. If you still get the wider line lengths, see if you have local pandoc settings/configs/modifications that change this.
4. If all else fails, submit this PR without html, and I will generate the html for you.

jon-edward:
Hey @aardappel, just removed the affected HTML files from my PR. Even with ensuring I had pandoc at 3.1.11.1, I still found that the line widths were slightly different from the hosted docs. Turns out I was running an even older version of pandoc than you, lol. 

Slight aside, it might be worth running doc generation and Pages upload on a GH Action to keep this in a consistent environment. I'll see if I can get this working, I think this should automate away a lot of this manual doc building and the version issues.

aardappel:
Your current PR deletes the htmls, and can't be merged like this.. please rebase and/or remove the html modifications from your PR.

As for using a GH action, that sounds nice, but we do want these html files part of the repo.. unless there is a way to automatically commit them afterwards or whatever.

jon-edward:
@aardappel, I think everything is good now. 

There is a way to do this, you can push changes from a GH action with the [Workflow permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions) setting.

aardappel:
Ok, thanks!

PR #303, opened 2024-02-19
Remove SDLSoundInit Emscripten directive

jon-edward:
From what I can tell, this Emscripten directive was introduced at commit 196c5cd7243927d45f89340df8e667096592f96a. Tested on Chrome and Firefox, I'm not experiencing the noise distortion that the comment is referencing anymore; I think it's safe to remove this restriction.


aardappel:
Cool.. yeah that was a long time ago, makes sense it be fixed by now :)

PR #306, opened 2024-03-09
Fixed some trivial typos (of ellipses).

WraithGlade:
Hello! I really love the design of your language so far and am looking forward to using it! 😎

Anyway, I found some minor typos and fixed them in the doc codebase.

I ran [the HTML regeneration batch script](https://github.com/aardappel/lobster/blob/master/docs/source/generate_html_docs_from_markdown.bat) too after the typo fixes. 

Should we do that for documentation fixes like this or should we not push the results of the regeneration scripts to the main repo? ⚠

I'm inclined to think you probably want it included, since that results in a more complete and consistent state.

Thank you for your time and efforts on this project and for reading my commit suggestion! 😁

### DIFF
--- a/docs/README_FIRST.html
+++ b/docs/README_FIRST.html
@@ -36,7 +36,7 @@ system and compile-time memory management with a very lightweight,
 friendly and terse syntax, by doing most of the heavy lifting for
 you.</p>
 <p>The homepage has more on <a href="http://strlen.com/lobster/">the
-“why” of Lobster</a>.</p>
+"why" of Lobster</a>.</p>
 <p>Chat on <a href="https://discord.gg/szJPYdX">Discord</a> or <a
 href="https://gitter.im/lobster_programming_language/community">Gitter</a>.</p>
 <p>This document is the starting point for all the Lobster
--- a/docs/README_FIRST.html
+++ b/docs/README_FIRST.html
@@ -11,8 +11,11 @@
     div.columns{display: flex; gap: min(4vw, 1.5em);}
     div.column{flex: auto; overflow-x: auto;}
     div.hanging-indent{margin-left: 1.5em; text-indent: -1.5em;}
-    ul.task-list{list-style: none;}
+    /* The extra [class] is a hack that increases specificity enough to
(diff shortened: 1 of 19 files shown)


aardappel:
Thanks! Yes, html changes should also be pushed, but in your case a lot of the changes in the diff are caused by having a different version of `pandoc` installed.. try installing the latest, then running again.

WraithGlade:
Ah, thanks for the info! I had forgotten that I haven't updated Pandoc a while.

I updated and redid the generation.

Is the new version good?

aardappel:
Ok, this looks better, but there are now cases where it replaces a `..` by a triple-dot single non-ascii character, I don't think thats an improvement.

WraithGlade:
Thanks for the helpful feedback again Mr. Oortmerssen! 🙂

I wasn't aware of what kinds of things you do/don't want in the docs and so this is good info to know.

It makes sense to avoid special unicode characters in order to prevent unexpected problems when programmers copy-paste things.

As such, I changed the Pandoc command to no longer perform any such special character conversions in the resulting HTML output.

However, consequentially a huge number of lines are now changed (almost all having to do with special variants of quotation marks and spaces), but by doing so many hidden special characters have now been removed, thus also removing some subtle copy-paste hazards for programmers.

It does seem to slightly change the appearance of the HTML though. 

I am not sure if it will be within your preferences, but I think so considering it is a technical/functional win (removes a hazard).

I found the info about how to fix this on [this Pandoc GitHub page](https://github.com/jgm/pandoc/issues/5707).

aardappel:
I guess this `-smart` option is an improvent. Only thing to do is to also add it to the `.sh` script.

aardappel:
Thanks :)

WraithGlade:
Thanks for accepting my pull request!

I tried looking if there was a way to only prevent the "..."s from getting converted but a quick internet search didn't turn anything up. Perhaps it is possible somehow though. I'm not that familiar with Pandoc. It seems like perhaps it doesn't offer that level of fine-grained control though.

Issue #305, opened 2024-03-05
Issue compiling Arm64 on M2


DeanHnter:
Hi,

Following the guide on https://aardappel.github.io/lobster/implementation.html i find if i try to compile using the command-line I receive the following error when on macos, M2 Sonoma Arm64

The command:

`xcodebuild -toolchain clang -configuration Release -target lobster`

Output:

```
ld: warning: ignoring file '/opt/homebrew/Cellar/llvm/17.0.5/lib/libunwind.1.0.dylib': found architecture 'arm64', required architecture 'x86_64'
ld: Undefined symbols:
  __Unwind_Resume, referenced from:
      flatbuffers::GenTextFile(flatbuffers::Parser const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in idl_gen_text.o
      flatbuffers::GenTextFile(flatbuffers::Parser const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in idl_gen_text.o
      flatbuffers::TextFileName(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) in idl_gen_text.o
      void flatbuffers::JsonPrinter::PrintScalar<unsigned char>(unsigned char, flatbuffers::Type const&, int) in idl_gen_text.o
      std::__1::__throw_length_error[abi:v160006](char const*) in idl_gen_text.o
      std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> flatbuffers::NumToString<int>(int) in idl_gen_text.o
      std::__1::basic_stringstream<char, std::__1::char_traits<char>, std::__1::allocator<char>>::basic_stringstream[abi:v160006]() in idl_gen_text.o
      ...
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

aardappel:
It seems to try to link the x86 version of `libunwind` even though you are on arm.. We don't refer to `libunwind` directly, so I am not sure where this is coming from..

DeanHnter:
This seems to have been an issue on my local machine only for some reason, on the CI it works fine so I will close this issue.

PR #307, opened 2024-03-10
Added Macos Matrix for Arm support

DeanHnter:
Hi,

After opening the Issue #305 as a test I modified the build pipeline and actually the artefact produced works on the M2 Arm64 however it requires a build matrix to support production of a Arm64 binary.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,10 +83,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-latest
-            artifact_name: Mac Lobster binary
+          - os: macos-12
+            artifact_name: Mac (Apple x64 Intel) Lobster binary
           - os: macos-14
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -79,7 +79,14 @@ jobs:
 
   build-mac:
     name: Build Mac
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: macos-latest
```
aardappel, line 86: maybe make this `macos-12` (the current equivalent), because at some point they'll move `latest` to mean `14` and then this will not work as intended.

Also maybe add `(x64)` or whatever in the description so its clear this is meant to be pinned to intel.

DeanHnter: Good points, thanks for the review! I Applied the suggested changes.

aardappel:
Thanks :)

PR #308, opened 2024-03-10
Fixed all warnings about floating point numbers written as integers throughout the samples directory, etc.

WraithGlade:
Fixed all warnings about floating point numbers written as integers throughout the samples directory, so that they don't distract new users.

Tested all of the files to ensure they continue to work.

Carefully made the corresponding changes to the `shooter_tutorial.md` doc, including the fragmentary code excerpts.

Regenerated the corresponding HTML docs.


aardappel:
Nice, thanks for the cleanup :)

WraithGlade:
Happy to help! This is a really cool programming language project! 😎

All the small tasteful choices really make the difference in how something feels, and its always nice to see projects/creators that genuinely care about that. 🏆

PR #310, opened 2024-03-12
Add Steam Lobby API

binji:
Also add lobbytest.lobster and update tut_net.lobster to use the new API

### DIFF
--- a/dev/src/steamworks.cpp
+++ b/dev/src/steamworks.cpp
@@ -35,6 +35,9 @@ struct SteamState {
     HSteamNetPollGroup poll_group = k_HSteamNetPollGroup_Invalid;
     vector<SteamPeer> peers;
     bool steamoverlayactive = false;
+    // Lobby Data
+    vector<CSteamID> joined_lobbies;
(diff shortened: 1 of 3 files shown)


binji:
just creating a PR so I can test the other builds

dev/src/steamworks.cpp
```diff
@@ -561,4 +770,250 @@ nfr("p2p_receive_messages", "", "", "S]S]", "receive messages from all"
...
+        if (steam) {
+            ok = steam->JoinLobby(SteamIDFromValue(steam_id));
+        }
+    #endif  // PLATFORM_STEAMWORKS
+    return Value(ok);
+});
+
+nfr("lobby_leave", "steam_id", "I", "B", "TODO", [](StackPtr &, VM &, Value &steam_id) {
+    auto ok = false;
+    #ifdef PLATFORM_STEAMWORKS
```
aardappel, line 795: This pattern is getting repetitive.. possibly a macro?

binji: Yeah, it's kind of a pain. I'll take a look

PR #319, opened 2024-07-31
fix typo in TextMate grammar

danilopedraza:
I didn't test the change, but I remember that highlighting of `let` statements  behave in a strange way.


aardappel:
this looks like an improvement regardless, thanks :)

PR #321, opened 2024-08-02
 Print builtin function returns and parse it in lsp

inferrna:
Print builtin function returns in format
`funcname(param: type) -> (typeA, typeB)`

also fixes lsp crash on lookup for user function (was failed to parse returns)

image
(above highlight is somewhat broken, but I didn't touch it)
image

### DIFF
--- a/dev/src/lobster/idents.h
+++ b/dev/src/lobster/idents.h
@@ -48,7 +48,7 @@ SlabAlloc *g_current_slaballoc;
             return g_current_slaballoc->dealloc_small(ptr); \
         }
 #else
-    #define USE_CURRENT_SLABALLOCATOR 
+    #define USE_CURRENT_SLABALLOCATOR
 #endif
 
 struct Ident : Named {
@@ -631,7 +631,7 @@ struct SymbolTable {
         for (auto tv  : typevars)         delete tv;
         for (auto su  : specudts)         delete su;
     }
-    
+
     bool MaybeNameSpace(string_view name) {
         return !current_namespace.empty() && name.find(".") == name.npos;
     }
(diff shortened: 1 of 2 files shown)


dev/src/lobster/idents.h
```diff
@@ -1380,6 +1380,14 @@ inline string Signature(const NativeFun &nf) {
         FormatArg(r, arg.name, i, arg.type);
     }
     r += ")";
+    if (nf.retvals.size() > 0){
```
aardappel, line 1383: space before `{`

dev/src/lobster/idents.h
```diff
@@ -1380,6 +1380,14 @@ inline string Signature(const NativeFun &nf) {
...
+        r += "->(";
```
aardappel, line 1384: spaces around `->`

dev/src/lobster/idents.h
```diff
@@ -1380,6 +1380,14 @@ inline string Signature(const NativeFun &nf) {
...
+        for (auto [i, retval] : enumerate(nf.retvals)) {
+            if (i) r += ", ";
+            r += TypeName(retval.type);
+        }
+        r += ")";
```
aardappel, line 1389: Since this is also used for printing certain error message, lets make it have the regular Lobster syntax that uses no `()`

aardappel (commented):
@ewoudje for the JS parts

dev/src/lobster/idents.h
inferrna: I just made it looks similar to definition of user defined functions, which was without spaces around "->" and with brackets around returning tuple. It seems like this code responsive for it
[dev/src/lobster/idents.h](https://github.com/aardappel/lobster/blob/72210a9af1861ae9099336a543d9b979dcb663db/dev/src/lobster/idents.h#L1406C1-L1426C2)

ewoudje:
LGTM

dev/src/lobster/idents.h
inferrna: Right now in release version of Lobster.
image
image
so should signature format of builtin function differs from user function?

aardappel: Good catch, they should all look like the definition, just changed the error message.

aardappel:
LGTM on the C++ part.. @ewoudje can you check the last TS changes?

ewoudje (approved):
Also LGTM 👍

PR #324, opened 2024-08-17
Highlight multiret

inferrna:
Fix this
Screenshot from 2024-08-17 13-11-48
to this
after

Also updated list of builtins to highlight with

```bash
lobster  --gen-builtins-json | jq  'map(.funcname)' - | awk '$1 ~ " *\"" {gsub(",", ""); print $1"|"}'|sort|uniq | xargs| tr -d \ 
```

so it became
after_after

@ewoudje behold it please


aardappel:
This LGTM

PR #325, opened 2024-08-20
Parse function signature with untyped parameter

inferrna:
Fixes lsp crash when parsing functions accepts parameter of any type ( like lobster_value_to_binary )
See also
https://github.com/aardappel/lobster/blob/a7b921a4f3fd4d661bb05a84a218cae5047cd93f/dev/src/lobster/idents.h#L1365C1-L1372C2


aardappel:
Thanks!

PR #330, opened 2024-09-01
more docs fixes

lesleyrs:
This should be the last of them, read through all of it except lang reference and builtin funcs. I appreciate the transparency in the docs rather than propaganda, although this language is quite hidden (only found it through your recent interview). 👍 💯

playervec line is taken from the shooter sample, it just wasn't included in the tutorial


aardappel:
Thanks!

Yeah I don't do any promoting of the lang, don't have time for that :)

`docs/builtin_functions_reference.html` is automatically generated by the compiler, so if there are any manual changes in there, those will eventually be overwritten.. any fixes would need to be made in the C++ code like you already did for `file.cpp`.

lesleyrs:
I didn't change the html manually this time, just ran the generate_builtin batchfile. For some reason the latest version of pandoc on winget changed formatting somewhat...

aardappel:
Ok, well maybe remove `docs/builtin_functions_reference.html` from the PR, I'll need to check whats going wrong there.

aardappel:
Thanks!

PR #342, opened 2024-12-28
Smallpt in OGL

Jaisiero:
Smallpt in a compute shader for learning purposes.
Happy holidays!


samples/demos/smallpt.lobster
```diff
@@ -121,8 +121,8 @@ def radiance(r:Ray, depth) -> float3:
             radiance(reflRay,depth)*Re+radiance(Ray { x, tdir },depth)*Tr
     return obj.e + f*temp
 
-let w = 64
```
aardappel, line 124: Lets not change this one, it will become way too slow :)

samples/demos/custom_shader_pt/custom_shader_pt.glsl
```diff
@@ -0,0 +1,37 @@
+COMPUTEFUNCTIONS
+    INCLUDE helper_functions.glsl
```
aardappel, line 2: Let's inline this file, will become easier to follow.

aardappel (commented):
cool, thanks :)

samples/demos/smallpt.lobster
Jaisiero: That's right. I forgot to change it back.

samples/demos/custom_shader_pt/custom_shader_pt.glsl
Jaisiero: Sounds good! I was trying to get used to the language/environment by using includes and so on.

aardappel:
Nice thanks!

PR #348, opened 2025-01-06
Fixed no type for freshly declared variable and nested typecheck

inferrna:
### Strong review needed - contains a lot of doubtful changes in code in doubtful language.

_especially look at parser.h:134_

- Fixed no type for freshly declared variable
- Nested typecheck for foo.bar.bas()
- Tastes better with #347

alpha is just declared but recognized correctly
image

method from Beta class (before there would be a collision with Alpha)
image

Detected type of nested class
image

and its method
image

field also detected
image


aardappel:
Yes, I'll need to review the C++ changes carefully. But you're making that hard by a) adding lots of whitespace changes where your editor has removed whitespace or you add blank lines, and b) introducing `ProcessDefinition` which means I can't see the diff with the existing code. Please remove these changes such that the diff only shows your actual changes to C++ code. Also make sure you fix any issues on CI related to your change before review.

PR #349, opened 2025-01-07
Fixes queries for lsp

inferrna:
- Fixed no type for freshly declared variable
- Nested typecheck for foo.bar.bas() constructions

This one is much better than #348 
Now there is only two possible breaking changes:
1. `LookupStruct` now search inside `gudttable`
2. `ProcessQuery` is now returns `bool`. Otherwise the call to `ProcessQuery` in `Compile` was always unreachable because of the `THROW_OR_ABORT` call in the end of `ProcessQuery` . So now it is required to call it manually after `ProcessQuery` returns `false`.

Also no more annoying spaces, etc. But I kept `ProcessDefinition` as a separate function because of its recursive nature.

still may get some collisions (below it should be `zk.d: int`)
image

### DIFF
--- a/dev/src/lobster/idents.h
+++ b/dev/src/lobster/idents.h
@@ -903,6 +903,12 @@ struct SymbolTable {
         }
         auto uit = gudts.find(name);
         if (uit != gudts.end()) return uit->second;
+        return nullptr;
+    }
+    GUDT *LookupStructQuery(string_view name) {
+        GUDT* res = LookupStruct(name);
+        if(res) return res;
+        //Try to search out of scope when doing query
         for (auto gudt = gudttable.rbegin(); gudt != gudttable.rend(); ++gudt) {
             if((*gudt)->name == name) {
                 return *gudt;
--- a/dev/src/lobster/idents.h
+++ b/dev/src/lobster/idents.h
@@ -903,6 +903,11 @@ struct SymbolTable {
         }
         auto uit = gudts.find(name);
         if (uit != gudts.end()) return uit->second;
+        for (auto gudt = gudttable.rbegin(); gudt != gudttable.rend(); ++gudt) {
+            if((*gudt)->name == name) {
+                return *gudt;
+            }
+        }
         return nullptr;
     }
 
--- a/dev/src/lobster/typecheck.h
+++ b/dev/src/lobster/typecheck.h
@@ -2333,15 +2333,15 @@ struct TypeChecker {
             }
         }
     }
-    std::optional<TypeRef> FindVarType(string_view ident, SubFunction *sf) {
+    TypeRef FindVarType(string_view ident, SubFunction *sf) {
         for (auto &vars : {sf->args, sf->locals, sf->freevars}) {
             for (auto &var : vars) {
-                if (var.sid->id->name == ident) {
+                if (var.sid->id->name == ident && !var.sid->type.Null()) {
                     return var.sid->type;
                 }
             }
         }
-        return std::nullopt;
+        return TypeRef(nullptr);
     }
 
     bool ProcessDefinition(GUDT *parent, string full_iden, SubFunction **sf) {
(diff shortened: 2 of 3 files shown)


dev/src/lobster/idents.h
```diff
@@ -903,6 +903,11 @@ struct SymbolTable {
         }
         auto uit = gudts.find(name);
         if (uit != gudts.end()) return uit->second;
+        for (auto gudt = gudttable.rbegin(); gudt != gudttable.rend(); ++gudt) {
```
aardappel, line 906: This would affect other users of `LookupStruct`. Looking up in `gudts` is scoped, in `gudttable` is not.

I'd suggest just making a a `LookupStructQuery` that first calls `LookupStruct` then checks `gudttable`.

dev/src/lobster/typecheck.h
```diff
@@ -2260,7 +2260,12 @@ struct TypeChecker {
         // Check if we need to do any lifetime adjustments.
         AdjustLifetime(n, recip, idents);
         // Check for queries.
-        if (query && query->qloc == n->line) ProcessQuery();
+        if (query) {
+            if ((reqret==0 && n->line == query->qloc) //reqret usually in the end of line
```
aardappel, line 2264: `reqret` indicates if there is a surrounding expression that requires a return value of the current expression, what is the effect of checking this? I don't understand the comment.

dev/src/lobster/typecheck.h
```diff
@@ -2260,7 +2260,12 @@ struct TypeChecker {
...
+             || (n->line.line > query->qloc.line && n->line.fileidx==query->qloc.fileidx)) { //If above missed
```
aardappel, line 2265: this will attempt to do the query in all lines following..

dev/src/lobster/typecheck.h
```diff
@@ -2328,47 +2333,96 @@ struct TypeChecker {
             }
         }
     }
+    std::optional<TypeRef> FindVarType(string_view ident, SubFunction *sf) {
```
aardappel, line 2336: A `TypeRef` has the ability to be null  built-in.

dev/src/lobster/typecheck.h
inferrna: So can I use `reqret==0` as an indicator that it is the last pass on this line?

dev/src/lobster/typecheck.h
inferrna: I'm a witness of "billion-dollar mistake" and prefer "no value" to be obvious.

dev/src/lobster/idents.h
inferrna: Ok, I also thought about something like this.

inferrna: See 250c228

dev/src/lobster/typecheck.h
inferrna: Fixed in ee52218

dev/src/lobster/typecheck.h
aardappel: I now finally understand what you're trying to do :)
`reqret==0` will work under most circumstances but it is very shaky, there may be exceptions.
Just using `>` for the line comparison seems more robust, and should be sufficient.

inferrna: But how then will be queried the last line…) I searched for anything that would be the last line indicator, but nothing found.
As for now, `reqret==0` is checked on the queried line, so there are two possible fails:
1. Early `reqret==0`. A little chance query to fail (found nothing).
2. No `reqret==0` at all on this line. May be a problem if line is the last one. Can we have the last line without `reqret==0` in it? If the answer is "No", then `reqret==0` is still a solution.

aardappel: Apologies for not responding sooner.

Ok I guess we can merge it for now with the reqret check, as there is no easy way to do the query on the last node that has a certain line number... but it really isn't a good check that may fail on occasion. We can think of a better solution later.

franca6:
Hopefully this is fixed cheers to everyone's input 🥰

PR #361, opened 2025-05-13
Extending support of more texture formats

Jaisiero:
Supporting a few more format for textures (image and samplers) mainly integer and unsigned integer formats.


aardappel:
Nice!

PR #362, opened 2025-05-16
Update lobster.vsix binary

inferrna:
Update compiled vscode extension file.


aardappel:
Thanks.. though at some point we should automate this somehow.

PR #364, opened 2025-05-19
Rolling back texture formats

Jaisiero:


aardappel:
Ah merged before I could say: this misses the `texture_format_` flag changes. Please add those.

PR #365, opened 2025-05-20
Added texture_format_32bit to enums

Jaisiero:


Jaisiero:
Is that **texture_format_32bit** what you mean?

Issue #187, opened 2022-07-18
Continue statement?


AntonBogun:
I could not find any description on a `continue` statement which can be seen in other languages, and it does not seem to be a reserved identifier inside loops. Is this intentional or was it just not implemented yet?

aardappel:
I guess I've simply never had a need for it. It could definitely be added.

482migs:
is this planned to be added, or is this just a dead thread? `continue` would be super helpful for lots of projects.

aardappel:
> is this planned to be added, or is this just a dead thread?

Neither. This is not an open source project with permanent staffing, so things get added when contributors feel like implementing things, not anything on a set priority or schedule.

Hjagu09:
I think the new guard statement could be used like a conditional continue... Correct me if I'm wrong but a separate continue statement shouldn't be necessary, you can always write ```guard false``` and it should be perfectly equal to an conventional ```continue```

aardappel:
`guard false` only shortcuts the current block, so wouldn't allow for the more common `if ...: continue`.
Also, `guard false` never executes the statements following it, so would be non-sensical.

tigercoding56:
> > is this planned to be added, or is this just a dead thread?
> 
> Neither. This is not an open source project with permanent staffing, so things get added when contributors feel like implementing things, not anything on a set priority or schedule.

i checked the doc folder,  sorry if this is a stupid question but  is there documentation on where keywords are implemented  i just found documentation about the lobster language itself

PR #379, opened 2025-11-03
add continue statement

danilopedraza:
I thought that it would be nice to have it after following the discussion in #187. Followed the `break` statement flow to implement it. Let me know what you think and if I missed something (I imagine that I did!).


docs/language_reference.html
```diff
@@ -1020,6 +1020,9 @@ <h3 id="break">break</h3>
 statement inside of them, that exits the enclosing loop. Alternatively,
 use <code>return</code> or <code>return from</code> (see above) for more
 complex cases.</p>
+<h3 id="continue">continue</h3>
```
aardappel: This is a generated file from `docs/source/language_reference.md`, you want to update it there.

aardappel (approved):

PR #384, opened 2025-12-05
modify syntax highlighting patterns for escaped characters in strings

danilopedraza:
The pattern for escaped characters was like this:
```regex
'(\\\\n|\\\\t|\\\\r|\\\\\"|\\\\'|\\\\x\\h{2}|\\w|\\\\|)'
```

Now is like this:
```regex
(\\\\n|\\\\t|\\\\r|\\\\\"|\\\\'|\\\\x\\h{2}|\\\\w|\\\\)
```
Previously, this happened: (`tests/misctest.lobster`, line 69)

image

With the new regex, and putting it as a subpattern of the string pattern, it looks like this:

image

I think I got all the existing cases right, but these things are tricky. Please let me know if I missed something.

### DIFF
--- a/dev/vscode-ext/syntaxes/lobster.tmLanguage.json
+++ b/dev/vscode-ext/syntaxes/lobster.tmLanguage.json
@@ -184,16 +184,24 @@
         {
           "name": "string.quoted.triple.lobster",
           "begin": "\"{3}",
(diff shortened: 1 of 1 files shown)


aardappel:
Nice, thanks for fixing!

PR #385, opened 2025-12-11
VSCode: Higlight enums same way as classes (changes classes too)

inferrna:
| Before | After |
| ------------- | ------------- |
| before  | after  |


aardappel:
Thanks!

Issue #398, opened 2026-02-05
Error at wrong line


MrJCraft:
```python
class Tjs:
      columns : int

var value : Tjs



// extra lines for test : Error shows up here
```
bug.lobster(9): error: missing declaration for value
correct error should be
bug.lobster(4): error: missing declaration for value

I checked it out as I already fixed a similar issue before.
I am making a pull request for the fix that I did for it. 
and the background needed info

aardappel:
This was fixed by the PR.

PR #399, opened 2026-02-05
Fix for Error at the wrong line

MrJCraft:
https://github.com/aardappel/lobster/issues/398

its on line 154 in dev/src/lobster/parser.h
Error("missing declaration for ", id->name);
doesn't pass the line information
the alternative function ErrorAt accepts a Node and not an Ident

I created a similar template to ErrorAt that takes a Line
I could have done it similarly to ErrorAt and passed the Ident
but passing a Line is more generic

I pass the line by pointer because lex.Warn and lex.Error expect pointers
I tried it with passing by value it worked but for consistency sake I used a pointer

### DIFF
--- a/dev/src/lobster/parser.h
+++ b/dev/src/lobster/parser.h
@@ -47,10 +47,6 @@ struct Parser {
         lex.Error(cat(args...), what ? &what->line : nullptr);
     }
 
-    template<typename... Ts> void ErrorAtLine(const Line *line, const Ts &...args) {
-        lex.Error(cat(args...), line);
-    }
-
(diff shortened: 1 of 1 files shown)


dev/src/lobster/parser.h
```diff
@@ -47,6 +47,10 @@ struct Parser {
         lex.Error(cat(args...), what ? &what->line : nullptr);
     }
 
+    template<typename... Ts> void ErrorAtLine(const Line *line, const Ts &...args) {
```
aardappel: There's already an `ErrorAt` that can be used with `d` as an argument in the caller?

aardappel:
Thanks!

PR #417, opened 2026-04-24
SDL3

binji:


aardappel:
2000 files changed woah.. thanks for pushing this :)

binji:
@aardappel This builds on all platforms now, and I've tested it on Linux and Windows and it works with all the samples/projects I've tried. I haven't tried any of the other platform builds yet, however.

Most of the port was pretty mechanical, just requiring some simple renames.

The biggest code changes were in `sdlsystem.cpp` and `sdlaudiosfxr.cpp`. The code for switching to fullscreen modes needed to be changed significantly. I tested it locally, but it may still have some bugs.

SDL_Mixer 3.0 has a completely different API, so the sound/music code needed to be completely rewritten. Thankfully, the new API is much more powerful and easy to use, so I took the opportunity to simplify some of the code here.

aardappel:
Very cool.. lets merge it, and address any remaining issues as they come.

PR #418, opened 2026-04-27
Make sure seconds are taken as unit

tobiolo:


aardappel:
Can you explain more? What exactly is the difference between the two?

We have other users of this code, so it shouldn't change the output

tobiolo:
In the documentation for the Lobster function it is explictly stated that the duration passed in as `time` are actually seconds:

> convert a time in *seconds* since 00:00:00 UTC, Thursday, 1 January 1970 into a string, using the same format 
> string syntax as POSIX strftime. If localtime is true, then the time will be displayed using the local timezone, 
> otherwise it will use UTC. Returns an empty string on error.

So without explicitly instructing that the value of the  `time` variable are seconds, it will just assume the default  tick unit. 

```
chrono::system_clock::time_point tp { chrono::system_clock::duration(time.ival()) };
```

This treats time.ival() as "ticks" of the system clock's native resolution.
The duration of a "tick" is platform-dependent. On many modern systems (like macOS or some Linux distros), system_clock::duration is measured in nanoseconds. On others, it might be microseconds or milliseconds.

```
chrono::system_clock::time_point tp { chrono::seconds(time.ival()) };
```

What this changes:
It explicitly treats `time.ival()` as seconds, regardless of what the underlying system clock uses. This is portable and predictable.

See https://timsong-cpp.github.io/cppwp/n4140/time.clock (object represents wall clock time by default, implementation-specific)

aardappel:
I guess on windows it is seconds? Since we've been using this without problems.

But thanks :)

tobiolo:
You are welcome!

tobiolo:
> I guess on windows it is seconds? Since we've been using this without problems.
> 
> But thanks :)

According to the [documentation](https://learn.microsoft.com/en-us/cpp/standard-library/system-clock-structure?view=msvc-170) of Microsoft, `system_clock::duration` represents a duration of time measured in units of 100 nanoseconds, not seconds in their own implementation.

tobiolo:
Thanks for the merge!

aardappel:
Ok, then I have no idea :)

tobiolo:
> Ok, then I have no idea :)

Ah you know your stuff for sure, no doubt about that. In C++, the details are really unforgivable, especially paired with divergence between vendors.

PR #421, opened 2026-05-04
Add audio recording API

binji:


binji:
Mostly wondering if the API names work for you. I was also considering `audio_input_*` or `mic_*`

aardappel:
So we have the existing `sound` namespace, but it is a namespace that doesn't require `sound.` to be present, and all our existing functions in it start with `sound_`, `play_`, `music_` and others.. so I'd say from that perspective, `mic_` would make most sense to me.

`audio_` would feel a bit like a double namespace in a way.

That naming is all a big mess of course, so this is what would fit with the existing mess :P

binji:
Pushed manually renaming `recording_*` to `mic_*`

PR #426, opened 2026-05-30
Load image

inferrna:
Adds load_image function. Sometimes it is needed to load an image rather than a texture. 
a

### DIFF
--- a/dev/src/builtins.cpp
+++ b/dev/src/builtins.cpp
@@ -19,6 +19,7 @@
 #include "lobster/wfc.h"
 #include "lobster/geom.h"
 
+#define STB_IMAGE_IMPLEMENTATION
 #include "stb/stb_image.h"
 
 namespace lobster {
--- a/dev/src/builtins.cpp
+++ b/dev/src/builtins.cpp
@@ -17,6 +17,9 @@
 #include "lobster/natreg.h"
 
 #include "lobster/wfc.h"
+#include "lobster/geom.h"
+
+#include "stb/stb_image.h"
 
 namespace lobster {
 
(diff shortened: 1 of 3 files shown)


aardappel:
This is cool but I'd suggest a few changes:

There already is a `create_texture` function that takes as input `F}:4]]` (i.e. `[[float4]]`) thus it would make sense that your function be the complement of that, i.e. `load_image` returns that same type, and thus `load_image` + `create_texture` is the same as `load_texture` except with the ability to modify the colors.

That also means putting `load_image` in `graphics.cpp`. It should probably call into a function in `gltexture.cpp` such that all image loading using STB is one place.

Maybe call it `load_image_data` so its even clearer you're not getting a usable image, but its data.

Colors in Lobster are uniformly represented with channels `0.0` to `1.0`, not ints to `255`. And you want the alpha for some images.

There's no need to return `height,width`, as that is obvious from the returned vectors.

The hard part is that we don't have a `TYPE_ELEM_VECTOR_OF_VECTOR_OF_FLOAT4` yet.. let me see if I can add that.

Btw, there was already `gl.read_texture` which allows texture data access but its slower and harder to use, so adding your function is still a good idea.

inferrna:
So, it seems to be two different functions: 
`load_image_data` returning `[int], ivec3` - current implementation
`load_image` returning [[float4]], which need some fancy logic: transform int to float, add alpha if there is not, transform single-channel with and without alpha to four channel

aardappel:
I just added support for `TYPE_ELEM_VECTOR_OF_FLOAT4` and `TYPE_ELEM_VECTOR_OF_VECTOR_OF_FLOAT4` which you'd need to implement my suggested changes.

My suggestion is to not have the `[int], ivec3` version at all.. it doesn't fit with the existing functions and how most Lobster functions handle color.

inferrna:
I did'n find a way to cast float4 as a Value. It must be 64 bit value, or pointer. Also, there is no builtin function returning array of any vector type.
So, it is needed some kind of PushVec but for arrays.

aardappel:
`float4` is an inline type, so it is 4 values, rather than a single one. You can definitely create vectors of those.

But I can imagine that is hard to figure out.. I can also add this function based on your code if you want.

aardappel:
I'm going to close this PR for now, let me know if you still need it and I will reopen.

Like I said, happy to add it for you if that is easier.

PR #427, opened 2026-06-27
feat: add scancode support for sdl input

tacf:
This PR adds support for hanling scancodes alongside with keys.

As a user of a modified non-standard layout, more often than not i find games that don't handle keys through scancodes which leads to having to do massing keybind changes to play a game. Scancodes bind to physical key codes instead of the char being "printed" by the software.

Also added support for using integer id for the scancode, but i'll be the first to say that it might not be as friendly or even useful having it there so, i can just drop it.


aardappel:
No game uses these as there is no sensible way to use these.

Either the game needs to use these in code directly, i.e. `gl.button("scancode 44")` but that is a bad idea since different keyboard manufacturers may use different scancodes. Maybe some scancodes are pretty stable but then you could have likely also used the existing key name.

Worse when the game offers user keybinding, as we don't want the user to have to know what `scancode 44` refers to.

So I see no reason this should be added.

As for the code, I don't understand why the function was introduced for lower case, it doesn't seem to be necessary. Also this codebase does not use camelCase it uses snake_case.

tacf:
"No game uses these". That's a big inaccurate statement. Lots of games and game engines support this. Just check Godot or Unity. Physical key can change from manufacturer to manufacturer but there's tipically an agreement or standard if you wanna see it that way. hence why they are supported by all major engines end frameworks. Just think that the French keyboard is AZERTY and not QWERTY and you have and entire community that will have to change ate least one major key.

As for the lower functions, it was already there I just moved it out to no have a hidden inline function in two places.

As mentioned the "scancode 44" does not seem correct although supported but "scancode W" should be the standard of anyone using it.

Games like Minecraft even use an approach where despite storing as sancodes they show the standard key in every interface so it's easier to visually match. But then use the proper key character in input boxes. Others approach it differently where they adjust everything even interfaces to display the matching character on that physical key.

aardappel:
Ok, looked into this, and it appears that SDL scancodes are not raw keyboard scancodes (which can differ), but normalized scancodes standardized on a US keyboard, so `scancode w` would represent whatever key is at the same location as `w` on a US keyboard.

That seems indeed nice to support.

I don't think we should support `scancode 44` unless absolutely necessary for some use case.

For a keybinding system (and in-game key indicators), we'll need need an additional function, one that turns `scancode w` into `z` for an AZERTY user, which would involve parsing `scancode w`, getting the scancode, calling `SDL_GetKeyFromScancode` on that, and turning it back into a name. That could be `gl.user_key` or something. You could add that, or if you don't want to, I can add it later.

Please also fix identifier case as noted.

It be good to add a list of all valid scancode names to the comment at the top of sdlsystem.cpp, though at some point this needs to be documented better.

tacf:
> I don't think we should support `scancode 44` unless absolutely necessary for some use case.

Agreed. I mostly left it as a example but that in specific i would consider a smell. What about the lower functions? I'm not the most versed in c++ std and libs it feels weird having a "standard" function there but i'll grant i understand why it wast like that in the first place.

> For a keybinding system (and in-game key indicators), we'll need need an additional function, one that turns `scancode w` into `z` for an AZERTY user, which would involve parsing `scancode w`, getting the scancode, calling `SDL_GetKeyFromScancode` on that, and turning it back into a name. That could be `gl.user_key` or something. You could add that, or if you don't want to, I can add it later.

Yes, that is indeed the standard use. I'll see if i can manage to do it cleanly. 

> It be good to add a list of all valid scancode names to the comment at the top of sdlsystem.cpp, though at some point this needs to be documented better.

Scancode list is a bit on the extensive side. Should it just refer to the SDL page/header instead? Like [this](https://wiki.libsdl.org/SDL3/SDL_Scancode) one.

aardappel:
> What about the lower functions? I'm not the most versed in c++ std and libs it feels weird having a "standard" function there but i'll grant i understand why it wast like that in the first place.

That's fine either way I guess.

> Scancode list is a bit on the extensive side. Should it just refer to the SDL page/header instead? Like [this](https://wiki.libsdl.org/SDL3/SDL_Scancode) one.

That doesn't list the strings. For example it has `SDL_SCANCODE_LSHIFT` but the string is `left shift` and there's no way to know that. So for the regular keycodes I dumped the results of the name function as a comment :)

tacf:
I pushed the change to remove the scancodes by ID and cleaned up a bit around that based on the linting definitions. Added a comment to clarify that there's a "built in" precedence (which is mostly irrelevant unless so weird decision is made on SDL side).

As for the comments with the full list of the supported scancodes that's where it gets a bit messy because it's a bit meaningless since they're "physically bound". Having the string/characters "supported" doesn't seem useful because you still don't know which exact physical key it represents. As an example, key "%" would seem equivalent to scancode "keypad %" but that key mostly only exists in larger 200+ keys keyboards.

https://zerico2005.github.io/SDL-Keycode-Scancode-Tool/ if you take a look here you can see the confusion that it would cause (and you can check the "%" example i mentioned).

Point being, without looking at that physical reference and a list of associated keycodes having a list in that plain format seems to be to do more harm than good. Unless you want me to generate an asscii art keyboard to add there :smile: 

PS.: Even you key list in the comment is missing a couple supported by SDL. Granted that they'll have little to no relevance for gaming, but users should be able to specify them either way.


As for the `gl.use_key`, i tried to figure out the structure to add that but was unable and i'm a bit afraid of any attempt just being a added burden on you to go back and forth on explaining things. If you'd be so kind as to do that part and tag me on the PR so then i can learn by example that would be great.

aardappel:
The code still does not conform to the existing code standard, but I can fix that later, and add `user_key`.

Thanks for the improvement :)

aardappel:
Here's the new `gl.user_key` and some cleanup: https://github.com/aardappel/lobster/commit/488159357ad0561f12daddc3e0536cb0fbd21a33

tacf:
Oh 'it was that easy'. Well now i know. Thank you (for doing basically the work :sweat_smile: ).

aardappel:
Well hey, you convinced me it was necessary :) I just added it to the default keybinds for WASD etc for my game, hopefully that will save some AZERTY and other people the need to rebind.

PR #450, opened 2026-09-22
Build Windows code paths with MinGW/GCC, not only MSVC

tobiolo:
Several Windows code paths are gated on `_MSC_VER` even though they depend on the platform, not the compiler. A MinGW build therefore falls through to the POSIX paths and fails. This came up while enabling Lobster in TreeSheets' MinGW cross-build (Linux → Windows, x86_64-w64-mingw32 GCC 16.2).

Changes:

- **`subprocess.h`**: use the Win32 implementation for `_WIN32` (otherwise it includes `spawn.h`, `sys/wait.h`, …). The `_MSC_VER` version checks are kept for MSVC. The `_DLL` check and the manual `_alloca` declaration are now MSVC-only, because MinGW defines `_DLL` as empty and `_alloca` is a macro for `__builtin_alloca`.
- **`slaballoc.h`**: use `_msize` for `_WIN32`, since `malloc_usable_size` is glibc-only.
- **`platform.cpp`**: with GCC on Windows, include the Windows headers before `stdafx.h`. Its `using namespace std` makes the Windows `byte` typedef ambiguous with `std::byte`. MSVC avoids this with `_HAS_STD_BYTE 0`, but libstdc++ has no such switch.

MSVC builds should be unaffected: every MSVC path either stays as it was or goes from `_MSC_VER` to `_WIN32`, which is also defined there. I couldn't build with MSVC myself.

Testing: TreeSheets compiles Lobster's sources into its own build. With these changes, its MinGW cross-build compiles and links (it also needs `-lsapi` for the text-to-speech GUIDs). Running `TreeSheets.exe -d` under Wine produces the same builtin-function reference as the native Linux build. I have not tried Lobster's own `dev/CMakeLists.txt` with MinGW.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aardappel:
The `process.cpp` changes seem problematic, since stdafx.h is universally always the first header included, and this breaks that. Worse, the block dealing with the windows include is now in there twice (on win32, windows.h will be included twice, I know it has guards, still, this is a messy fix).

I will look into improving the namespace situation independently, that may help.

tobiolo:
Thanks for your comments. Should I then wait until you fix the namespace situation?

aardappel:
Whoops, didn't mean to close it.

I made https://github.com/aardappel/lobster/commit/5f17315897ea08db675700fd4665b7ce24a80a11 on top of your change, so now std::byte is no longer included, and thus I could delete the platform.cpp hack.

Do you need another release tag to be able to pin TreeSheets?

tobiolo:
Ah perfect! Thanks for the teamwork here!
I am checking with the latest commit currently. Please stand by :-)

tobiolo:
Ok, I prepared TreeSheets with https://github.com/aardappel/treesheets/commit/ee335065454fe974f45544f9b04906c3d86c47d1. From my side you are ready to go :-)

aardappel:
tag `v2026.8` pushed.

tobiolo:
A big thank you, or merci as they say in Switzerland :-)

